PR #4, opened 2021-07-21
Add some utility bash scripts

joelstobart-moj:


run-local.sh
```diff
@@ -0,0 +1,39 @@
+#
+# This script is used to run the Create and Vary a licence API locally, to interact with
```
timharrison-moj, line 2: title text might need a name change?

timharrison-moj (approved):
LGTM- just one comment

run-local.sh
joelstobart-moj: Good point well made :)

timharrison-moj (approved):
LGTM

PR #5, opened 2021-07-21
Files in values and values-dev

joelstobart-moj:


src/main/resources/application-dev.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+# Most values are inherited from - application.yml - except the following:
+
+spring:
+  security:
+    oauth2:
+      resourceserver:
+        jwt:
+          public-key-location: classpath:local-public-key.pub
+
```
timharrison-moj, line 15: Probably need a copy of this file `local-public-key.pub` from CVL? It goes into `src/main/resources`

timharrison-moj (commented):
LGTM - one comment only. Can go in now or later, just wondering if it would start up with that missing though?

src/main/resources/application-dev.yaml
joelstobart-moj: I've addressed that one, also fixed an indenting issue in the application.yml

timharrison-moj (approved):
LGTM

PR #11, opened 2021-07-22
CRS-111

joelstobart-moj:
Add an Offender object and the related test
Ignore the Mac file DS_Store files

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/Offender.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/Offender.kt
@@ -1,9 +1,6 @@
 package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.model
 
 class Offender(
-  private var referenceField: String,
-  private var nameField: String
-) {
-  val name get() = nameField
-  val reference get() = referenceField
-}
+  val reference: String,
+  val name: String
+)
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/Offender.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.model
+
+class Offender(
+  private var referenceField: String,
+  private var nameField: String
+) {
+  val name get() = nameField
+  val reference get() = referenceField
```
timharrison-moj, line 8: Does it need a class body at all? Will the name / reference fields ever be altered after an offender object is constructed? Just a question as I don't know how this might be used or evolve. It's fine for now though could simply be `class Offender (val reference: String = "", val name: String = "")` ?

timharrison-moj (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/Offender.kt
joelstobart-moj: If I did that it created the Setters and I didn't want public setters (just public getters)

yasinmustafa (approved):
LGTM

PR #14, opened 2021-07-23
CRS-100 add postgres to docker-compose and set up db name and user details for local running

yasinmustafa:


joelstobart-moj (approved):
LGTM

PR #34, opened 2021-08-09
CRS-172 Integrate the prison-api with the service and implement a test call

yasinmustafa:


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/TestService.kt
```diff
@@ -5,9 +5,15 @@ import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.model.TestData
 import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.repository.TestRepository
 
 @Service
-class TestService(private val testRepository: TestRepository) {
+class TestService(
```
yasinmustafa, line 8: This whole test service and associated Test classes will be removed once the service is a bit more mature.. just being used to aid integration testing

joelstobart-moj (approved):
LGTM

PR #37, opened 2021-08-10
CRS-170 Add endpoint to pass a JSON string to test our calculation engine - will be used for testing purposes only

yasinmustafa:
This adds another test endpoint so that the calculation engine can be tested via postman / the front end

Will probably only be a temporary endpoint to aid testing by the wider team until our app is more mature

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/TestService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/TestService.kt
@@ -1,8 +1,19 @@
 package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.service
 
+import com.squareup.moshi.FromJson
+import com.squareup.moshi.JsonAdapter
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/TestService.kt
```diff
@@ -1,19 +1,98 @@
...
+
   fun getTestData(): List<TestData> {
     val prisoner = prisonService.getOffenderDetail("A1234AA")
     val testData = testRepository.findAll().map { transform(it) }.toMutableList()
-    testData.add(TestData(prisoner.offenderNo, prisoner.bookingId))
+    testData.add(TestData(prisoner.offenderNo, prisoner.bookingId.toString()))
     return testData
   }
+
+  fun jsonToBooking(json: String): Booking {
```
yasinmustafa, line 39: This `jsonToBooking` method has been copied from `JsonTranformer` which is in the test scope

joelstobart-moj (approved):

PR #38, opened 2021-08-11
CRS-154 Allow access to our API endpoints from any IP i.e. so we can make direct API calls without having to be on the VPN

yasinmustafa:


joelstobart-moj (approved):
LGTM

PR #42, opened 2021-08-13
CRS-139

joelstobart-moj:
PSI Example 28
- Addition of test expectation input and output.
- Addition of intermediate calculation values
- Rename of existing values for clarity
- Amendment to LED and NPD calculations. Was previously incorrect for 2 reasons:
  -  The calculation was always returning zero days
  -  The calculation was using the sentence length before days were subtracted


timharrison-moj (approved):
LGTM

PR #48, opened 2021-08-19
CRS-185 Add enpoint to calculate release dates for a prisonerId + lookup all the details to populate booking object to run the calculation

yasinmustafa:
TODO's

- Add integration test for new endpoint (requires adding mock endpoints for  prison api - will do as a separate task)
- Add screen to the front end side that that shows the results for a prisoner


util_scripts/test-api-locally.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
yasinmustafa, line 1: this utility script allows to test the locally running service via direct api calls

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/PrisonerDetails.kt
```diff
@@ -1,6 +1,11 @@
 package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.model
 
+import java.time.LocalDate
+
 data class PrisonerDetails(
```
yasinmustafa, line 5: These 3 data classes (PrisonerDetails, SentenceAdjustments and SentenceTerm) represent responses from the prisonaApi

joelstobart-moj:
LGTM

joelstobart-moj (approved):

PR #57, opened 2021-09-09
CRS-280 Create data model for calcultion-request and response + create JPA classes and fix int tests

yasinmustafa:
Also:
- Changed alerts channel to farsight-alerts
- Added a upgrade command into Dockerfile as advised by devops

TODO
- Populate calculation requests and results to the new tables
- add tests to check the population

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/entity/CalculationOutcome.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/entity/CalculationOutcome.kt
@@ -1,26 +1,27 @@
 package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.entity
 
-import com.vladmihalcea.hibernate.type.json.JsonBinaryType
-import org.hibernate.annotations.TypeDef
-import org.hibernate.annotations.TypeDefs
 import java.time.LocalDateTime
-import java.util.UUID
+import java.util.*
 import javax.persistence.Entity
 import javax.persistence.GeneratedValue
 import javax.persistence.GenerationType
 import javax.persistence.Id
 import javax.persistence.Table
+import javax.validation.constraints.NotNull
 
 @Entity
 @Table
-@TypeDefs(
-  TypeDef(name = "jsonb", typeClass = JsonBinaryType::class)
-)
 data class CalculationOutcome(
   @Id
   @GeneratedValue(strategy = GenerationType.IDENTITY)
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/entity/CalculationOutcome.kt
```diff
@@ -0,0 +1,26 @@
...
+import javax.persistence.GeneratedValue
+import javax.persistence.GenerationType
+import javax.persistence.Id
+import javax.persistence.Table
+
+@Entity
+@Table
+@TypeDefs(
+  TypeDef(name = "jsonb", typeClass = JsonBinaryType::class)
+)
```
timharrison-moj, line 18: As there is no JSONB types in this entity it probably doesn't need the @Typedefs?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/entity/CalculationRequest.kt
```diff
@@ -0,0 +1,37 @@
...
+@Table
+@TypeDefs(
+  TypeDef(name = "json", typeClass = JsonType::class)
+)
+data class CalculationRequest(
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  val id: Long = -1,
+  val calculationReference: UUID = UUID.randomUUID(),
+  val prisonerId: String = "",
```
timharrison-moj, line 28: Is null better than empty strings "" for default values? Or are the columns not nullable?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/entity/CalculationOutcome.kt
yasinmustafa: good spot - copy/paste error

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/entity/CalculationRequest.kt
yasinmustafa: They are actually not nullable.. will annotate with @NotNull

timharrison-moj (dismissed):
LGTM

timharrison-moj (approved):
LGTM

PR #61, opened 2021-09-28
CRS-273 Initial work for publishing an event to the domain-events topic to notify of a confirmed release dates change

yasinmustafa:
This was mainly copied from another project that publishes events to the domain-events topic - see the following PR for their code
https://github.com/ministryofjustice/hmpps-restricted-patients-api/pull/46


joelstobart-moj (commented):
LGTM

That's a big'o'config

joelstobart-moj (approved):

PR #63, opened 2021-09-29
CRS-437 Rename startedAt field to committedAt, remove endedAt field + use new version of endpoint to populate committedAt

yasinmustafa:
In addition:
- added integration tests to the circle build pipeline 
- added missing env vars for int tests to build (related to event queues)

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -103,7 +103,32 @@ management:
 domain-events-sqs:
   provider: aws
   endpoint.region: eu-west-2
+  aws:
+    access:
+      key:
+        id: access_key_id
+    secret:
+      access:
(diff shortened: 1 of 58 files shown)


src/main/resources/application.yml
```diff
@@ -103,7 +103,32 @@ management:
 domain-events-sqs:
   provider: aws
   endpoint.region: eu-west-2
+  aws:
```
yasinmustafa, line 106: Added this to allow the int tests to work - as the config beans cant be created without these values defined. However, these values will not be used in a deployed environment (like dev) as they are overridden by the actual values

joelstobart-moj (approved):
LGTM

PR #64, opened 2021-09-30
CRS-441 Create a POST endpoint to confirm a calculation of release dates and publish an event to domain-topic 

yasinmustafa:


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/CalculationService.kt
```diff
@@ -59,9 +59,6 @@ class CalculationService(
       calculationOutcomeRepository.save(transform(calculationRequest, it.key, it.value))
     }
 
-    if (calculationStatus == CONFIRMED) {
-      domainEventPublisher.publishReleaseDateChange(calculationRequest.prisonerId, calculationRequest.bookingId)
-    }
     return bookingCalculation
```
yasinmustafa, line 62: Ive moved the publishing of the event up one level - as we have a specific endpoint for confirming a calculation. Moving it up one level means we dont need the if condition

joelstobart-moj (approved):
LGTM

PR #68, opened 2021-10-12
Crs 445 scenario 2

joelstobart-moj:


yasinmustafa (approved):
LGTM

build.gradle.kts
```diff
@@ -92,4 +92,9 @@ tasks {
       jvmTarget = "16"
     }
   }
+  compileTestKotlin {
+    kotlinOptions {
+      jvmTarget = "16"
+    }
+  }
```
mikehalmamoj, line 99: This would apply to both main and test code: https://github.com/ministryofjustice/hmpps-template-kotlin/blob/main/build.gradle.kts#L19

PR #78, opened 2021-11-02
CRS-514 remove test_data table and associated code. Re-number flyway versions so we have a cleaner baseline for private beta

yasinmustafa:


src/test/resources/test_data/load-base-data.sql
```diff
@@ -1,3 +1 @@
-INSERT INTO test_data (key, value) values ('A', 'AAAAAA');
-INSERT INTO test_data (key, value) values ('B', 'BBBBBB');
-INSERT INTO test_data (key, value) values ('C', 'CCCCCC');
\ No newline at end of file
+insert into calculation_request(calculation_reference, prisoner_id, booking_id, calculation_status, calculated_at, calculated_by_username, input_data) values ('AABBCCDD', 'A1234AB', 12345, 'preliminary', now(), 'int_test_user', '{}');
```
ldlharper, line 1: insert into should be uppercase.

ldlharper (approved):
LGTM

src/test/resources/test_data/load-base-data.sql
yasinmustafa: thanks - will update next time as I  want to check if the deployment works

PR #82, opened 2021-11-05
CRS-592 fixing display of duration.

ldlharper:
Removing 0's from display of durations and pluralising the units.


joelstobart-moj (approved):
LGTM

PR #85, opened 2021-11-09
CRS-540 Fixing calcuation breakdown.

ldlharper:
Calculation test examples I've done in the past had the wrong consecutiveUUIDs. The consecutiveUUIDs should reference the sentence that is consecutive to, I had it the other way around.


joelstobart-moj (approved):
LGTM

PR #87, opened 2021-11-10
CRS-543 Allow sentences to be combined if tused and hdced are not in …

ldlharper:
…both sentences.


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/Sentence.kt
```diff
@@ -102,6 +102,17 @@ data class Sentence(
       )
   }
 
+  private fun sameSentenceTypesWithout(secondSentence: Sentence, vararg sentenceTypes: SentenceType): Boolean {
+    return this.sentenceTypesWithout(*sentenceTypes)
+      .containsAll(secondSentence.sentenceTypesWithout(*sentenceTypes)) &&
```
mikehalmamoj, line 107: A minor thing - but Kotlin has a function `.and(...)` which might make this even more readable than it already is.

mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/Sentence.kt
ldlharper: Ah I didn't know that thanks!

PR #88, opened 2021-11-11
CRS-193 Change calculation algorithm to cater for Example 39 from the PSI policy document

yasinmustafa:
Extra changes:

- (FIXED) This change exposed a bug in that we were deducting a day at the end of the calculation.. we should have been doing it before
A small explanation of the problem -
31 May + 1 month - 1 day = 29 June :x: (the PSI doesn't do it this way)
31 May - 1 day + 1 month = 30 June :heavy_check_mark: (edited) 

- Also removed the exception that was thrown if the 'sentence types' (badly named) for consecutive sentences didn't match. This was originally added as a belt and braces but is not required and costly to maintain.

TODO
In a another ticket rename the SentenceType class to ReleaseDateType or similar


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/Sentence.kt
```diff
@@ -3,6 +3,10 @@ package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.model
 import com.fasterxml.jackson.annotation.JsonIgnore
 import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.enumerations.SentenceIdentificationTrack
 import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.enumerations.SentenceType
+import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.enumerations.SentenceType.ARD
```
yasinmustafa, line 6: these are not really SentenceTypes.. will create a tech debt ticket to rename the class

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/Sentence.kt
```diff
@@ -86,32 +90,6 @@ data class Sentence(
       "Top-up Expiry Date (Post Sentence Supervision PSS)\t:\t" +
       "${sentenceCalculation.topUpSupervisionDate?.format(formatter)}\n"
   }
-  fun canMergeConsecutivelyWith(secondSentence: Sentence): Boolean {
```
yasinmustafa, line 89: All these methods are unused now that we dont throw the CannotMergeSentencesException

PR #90, opened 2021-11-12
CRS-550 tempoary change just to test if the api call works in dev.. will be followed up by a proper PR

yasinmustafa:


timharrison-moj (approved):
LGTM - experimental change to try in DEV.

PR #91, opened 2021-11-15
CRS-550 make call to prison-api to write release dates back to NOMIS

yasinmustafa:
This currently only writes CRD, SED and LED. The prison-api endpoint is still WIP and wil add tthe other dates as we progress

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -28,7 +28,7 @@ spring:
             client-id: ${system.client.id}
             client-secret: ${system.client.secret}
             authorization-grant-type: client_credentials
-            scope: read
+            scope: read,write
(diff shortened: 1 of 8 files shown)


src/main/resources/application.yml
```diff
@@ -28,7 +28,7 @@ spring:
             client-id: ${system.client.id}
             client-secret: ${system.client.secret}
             authorization-grant-type: client_credentials
-            scope: read
+            scope: read,write
```
yasinmustafa, line 31: Needed to add write scope here - otherwise the POST to prison-api failed with a 403 Forbidden

joelstobart-moj (approved):
LGTM

PR #94, opened 2021-11-17
CRS-571 Do not run the detekt static analysis tool as part of the standard build process + remove suppressions

yasinmustafa:


joelstobart-moj (approved):
LGTM

PR #95, opened 2021-11-18
CRS-194 PSI Example 40

ldlharper:
Calculation of a consecutive sentence made up of an old style schedule 15 sentence and a new style sentence.

Calculation of parole dates is done after the notional CRD, calculated from the new style sentence.


yasinmustafa (approved):
LGTM

PR #101, opened 2021-11-26
CRS-602 Use the users token from the request when making prison-api calls. Also remove the use of SYSTEM_CLIENT_ID and SYSTEM_CLIENT_SECRET

yasinmustafa:


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/config/UserContext.kt
```diff
@@ -0,0 +1,14 @@
+package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.config
+
+import org.springframework.stereotype.Component
+
+@Component
+object UserContext {
+  private val authToken = ThreadLocal<String>()
```
timharrison-moj, line 7: Should this have @RequestScope?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/config/UserContext.kt
```diff
@@ -0,0 +1,14 @@
...
+  fun setAuthToken(aToken: String) {
+    authToken.set(aToken)
+  }
+  fun getAuthToken(): String {
+    return authToken.get()
```
timharrison-moj, line 12: Don't think you need getters and setters?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/config/UserContextFilter.kt
```diff
@@ -0,0 +1,28 @@
...
+@Component
+@Order(1)
+class UserContextFilter : Filter {
+  @Throws(IOException::class, ServletException::class)
+  override fun doFilter(servletRequest: ServletRequest, servletResponse: ServletResponse, filterChain: FilterChain) {
+    val httpServletRequest = servletRequest as HttpServletRequest
+    val authToken = httpServletRequest.getHeader(HttpHeaders.AUTHORIZATION)
+    UserContext.setAuthToken(authToken)
+    filterChain.doFilter(httpServletRequest, servletResponse)
+  }
```
timharrison-moj, line 24: Maybe pull the username from the request too? Or is this done elsewhere?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/config/WebClientConfiguration.kt
```diff
@@ -20,15 +18,22 @@ class WebClientConfiguration(
...
+      .filter(addAuthHeaderFilterFunction())
       .build()
   }
 
+  private fun addAuthHeaderFilterFunction(): ExchangeFilterFunction {
+    return ExchangeFilterFunction { request: ClientRequest, next: ExchangeFunction ->
+      val filtered = ClientRequest.from(request)
+        .header(HttpHeaders.AUTHORIZATION, UserContext.getAuthToken())
+        .build()
+      next.exchange(filtered)
```
timharrison-moj, line 33: Ah. Username is already encoded in the token, so receiving APIs already get it.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/config/WebClientConfiguration.kt
```diff
@@ -43,17 +48,4 @@ class WebClientConfiguration(
   fun bankHolidayApiWebClient(): WebClient {
     return webClientBuilder.baseUrl(bankHolidayApiUrl).build()
   }
```
timharrison-moj, line 50: There's something else here. I think the `authorizedClientManager` might also check the validity of the token received in the request - it has a URL to the oauth2 provider (auth) to get its public key to be able to do this. What you have it a direct pass-thru of received token out into the request headers, with no check that it is valid. This might be ok - as the prison API will also do this check, but it feels like the checking should still be done?

src/main/resources/application.yml
```diff
@@ -21,17 +21,6 @@ spring:
...
-        registration:
-          prison-api:
-            provider: hmpps-auth
-            client-id: ${system.client.id}
-            client-secret: ${system.client.secret}
-            authorization-grant-type: client_credentials
-            scope: read,write
-        provider:
-          hmpps-auth:
-            token-uri: ${hmpps.auth.url}/oauth/token
```
timharrison-moj, line 24: This last line above - is used to get the public key of the auth provider (hmpps-auth) so the validity of received tokens can be checked.

timharrison-moj (commented):
Looks ok - but I think there is a piece missing - checking the validity of the received user token. Have a read into the `authorizedClientManager` to see whether the client-credentials part can be pulled out whilst retaining the token-checking features? I also wonder whether other services are passing through the token without a check?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/config/UserContext.kt
yasinmustafa: You can't add @RequestScope to `objects` - it causes an error. I've copied this from whereabouts; however I believe this is inherently 'request scoped' because of ThreadLocal

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/config/UserContext.kt
yasinmustafa: In this case they're not just getters and setters.. they are returning a value inside a ThreadLocal object

src/main/resources/application.yml
yasinmustafa: Added the provider section back in

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/config/WebClientConfiguration.kt
yasinmustafa: Will speak to Pete about how the validating of tokens work

yasinmustafa: Have spoken to Simon Mitchell from the tech team - he's had a look and is happy with the way it's been configured. He agreed that the authorizationManager isn't related to token verification. He will take this off as a separate task and discuss with the tech team to clarify exactly what happens. In the meantime we can press on.

He also recommended adding extra app-insight related config. Refer to https://github.com/ministryofjustice/hmpps-manage-users-api/tree/main/src/main/kotlin/uk/gov/justice/digital/hmpps/manageusersapi/config
Add 
1. ClientTrackingConfiguration.kt
2. ApplicationInsightsConfiguration.kt
Raised ticket CRS-606

src/test/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/integration/WorkingDayIntTest.kt
```diff
@@ -0,0 +1,59 @@
+package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.integration
```
yasinmustafa, line 1: This replaced the WorkingDayServiceInt test - because WorkingDayServiceInt was meant to be an int test but was bypassing the controller  and going straight into the service.

joelstobart-moj (approved):
LGTM

PR #102, opened 2021-11-29
CRS-466 unsupported sentence errors.

ldlharper:
Validate unsupported sentences.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/SupportedSentence.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/SupportedSentence.kt
@@ -1,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.model
+
+import java.time.LocalDate
+
+data class SupportedSentence(
+  val sentenceType: String,
+  val sentenceAtCondition: (LocalDate) -> Boolean = { true },
+  val offenceDateCondition: (LocalDate) -> Boolean = { true },
+)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/SupportedSentence.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/SupportedSentence.kt
@@ -3,7 +3,7 @@ package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.model
 import java.time.LocalDate
 
 data class SupportedSentence(
-  val sentenceType: String,
+  val sentenceType: SentenceType,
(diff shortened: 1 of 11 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/SupportedSentence.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.model
+
+import java.time.LocalDate
+
+data class SupportedSentence(
```
yasinmustafa, line 5: Not sure what the sentenceAtCondition and offenceDateCondition is for? Cant see how it's being used?

Can't this just be an enum? and then when you check if it's a valid SentenceType just check the list of enum values

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/SentenceValidationService.kt
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }
+
+  private fun canSentenceBeCalculated(sentenceAndOffence: SentenceAndOffences): Boolean {
+    return supportedSentences.any { it.sentenceType == sentenceAndOffence.sentenceCalculationType }
+  }
+
+  companion object {
+    private val supportedSentences: List<SupportedSentence> = listOf(
+      SupportedSentence("ADIMP", offenceDateCondition = { it.isAfterOrEqualTo(ImportantDates.CJA_DATE) }),
```
yasinmustafa, line 27: See my comment about using an enum for SupportedSentence and generating a list from the enum.. 
A bit tidier codewise if instead of writing 
ImportantDates.CJA_DATE
you just write
`CJA_DATE`
and import it

yasinmustafa (dismissed):
A couple of comments - happy to defer

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/SupportedSentence.kt
ldlharper: So they're not currently being used. Originally this ticket validated the sentence type and that the sentencedAt and offenceDate are valid for the given sentence type. However the second part was descope and will be covered in another ticket, I didn't want to delete it though because we'll need it later. I'll add a code comment.

I've added an enum for the SentenceType as well.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/SentenceValidationService.kt
ldlharper: Done thanks.

yasinmustafa (approved):
LGTM

PR #105, opened 2021-11-29
CRS-516 remove client id and client secret from values.yaml - had already been removed from elsewhere

yasinmustafa:
These keys are not required by our service any more - as we pass through the users token and this service does not require any extra roles


joelstobart-moj (approved):
LGTM

PR #106, opened 2021-11-30
CRS-610 Do not propogate error if publishing the event fails - simply log it

yasinmustafa:
Publishing the event doesn't constitute a failure at the moment because we are writing back to NOMIS using a POST endpoint.
Eventually the event will be used to write back to NOMIS in an async fashion and then this will need refactoring


joelstobart-moj (approved):
LGTM

PR #116, opened 2021-12-14
Crs 637 overlapping logic fix

ldlharper:
There is a bit of work on this PR. It started because there was an exception on a test case Santosh was looking at, this showed two bugs

1. There was a bug with the old overlap code, it was checking that EVERY sentence overlapped with EVERY other sentence.
2. There was a bug with the served ada code where by if sentence A started after and ended before sentence B it would consider this a gap to be filled by ADAs. This logic has been refactored significantly and moved. I think it fits in the extraction service and I've added a code comment to say what my intention would be in the future for this method. I.e. dealing with remanded for an upcoming court case and license recalls.


yasinmustafa (approved):
LGTM

PR #135, opened 2022-01-13
CRS-666 new JSON shape.

ldlharper:
New JSON shape. This will change once when we come to work on the next few stories, but this is a start. Any feedback let me know!


joelstobart-moj (approved):
LGTM

PR #141, opened 2022-01-26
Crs 685 Sentence level adjustments: tagged bail

ldlharper:
The calculations mostly work already thanks to the work done for remand. The only change needed was dealing with a release in the middle of a booking the same way as remand. Added test examples.


yasinmustafa (approved):
LGTM

PR #142, opened 2022-01-26
CRS-629 Add functionality to the algorithm to add granular level information about how a calc breakdown was achieved for TUSED and HDCED

yasinmustafa:
This includes a rules based approach to identify how a calculation was achieved (by release date type)

At the moment, the rules are only used for HDCED and TUSED - but will be expanded out to other Release Date types. This approach means the UI doesnt have to do any derivation and can use the rules to determine how the calculation was achieved.


ldlharper (approved):
LGTM

PR #152, opened 2022-02-03
CRS-712 add release and expiry dates with adjustment information to the calculation breakdown endpoint

yasinmustafa:


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/BookingCalculation.kt
```diff
@@ -10,7 +10,5 @@ data class BookingCalculation(
   var calculationRequestId: Long = -1,
   var effectiveSentenceLength: Period = Period.ZERO,
   @JsonIgnore
-  var daysAwardedServed: Long = 0,
```
yasinmustafa, line 13: This was never actually being set.. so removed it

joelstobart-moj (approved):
LGTM

PR #157, opened 2022-02-07
CRS-735 Remove the prisoners name from the input data that is used for the calculation - the prisoners name isn't used by the algorithm

yasinmustafa:


joelstobart-moj (approved):
LGTM

PR #190, opened 2022-03-24
CRS-810 adding the calc status to the calculation api. Also refactori…

ldlharper:
…ng for more immutablilty.


joelstobart-moj (approved):
LGTM

PR #233, opened 2022-06-22
CRS-924 Some pcsc cleanup

ldlharper:
Now the frontend journey is complete we can remove some of this old code for the old SDS+ journey.


joelstobart-moj (approved):
LGTM

PR #236, opened 2022-06-23
Update to the new Security DSL syntax

joelstobart-moj:
Hopefully, you can see that the lines in these files match. I just wanted to remove a deprecation warning.


ldlharper (approved):
LGTM

PR #240, opened 2022-06-24
Updating config to match NOMIS

ldlharper:
NOMIS are using the first of April as the pcsc commencement date for testing. Updating to match that so we can compare calculations.


joelstobart-moj (approved):
LGTM

PR #292, opened 2022-09-28
Crs 1052 POC for GDS linting in circle ci pipeline.

ldlharper:
This is a proof of concept for adding GDS linting into the pipeline.

Added jobs:

1. generate the api doc. This uses the gradle plugin https://github.com/springdoc/springdoc-openapi-gradle-plugin
The main issue I had here is that in order to generate the openapi file the plugin needs to spin up the spring boot service. So I've copied the configuration for running spring tests by running the service against h2. The ugly part of this is including h2 as a `implementation` rather than `testImplementation`. This is done with a gradle property so that its not included on the main build. I'm open to any other suggestions, I couldn't figure out how to run the app with test dependencies.

2. lint the api doc. This is configured in the `.spectral` file.

As expected the build fails and circle ci displays the rules that are failing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,6 +50,8 @@ jobs:
           command: ./gradlew -Pdocs generateOpenApiDocs
       - run:
           command: mkdir ~/docs
+      - run:
+          command: cp build/openapi.json ~/docs
       - save_cache:
           paths:
             - ~/.gradle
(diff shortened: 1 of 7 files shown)


joelstobart-moj (approved):
LGTM

Mjwillis (approved):
Looks Spot on!

joelstobart-moj (approved):
LGTM

PR #303, opened 2022-10-14
CRS-1068 Upgrade generic service, hmpps orb and some libraries

yasinmustafa:


.trivyignore
```diff
@@ -0,0 +1,22 @@
+# WARNING - THIS FILE WAS GENERATED BY THE dps-gradle-spring-boot GRADLE PLUGIN
+#           AND ANY MANUAL CHANGES WILL BE OVERRIDDEN ON YOUR NEXT BUILD.
+#
```
joelstobart-moj, line 3: I don't think this file should be committed?

joelstobart-moj (approved):
LGTM

PR #306, opened 2022-10-26
CRS-1070 Chenge the date range check for applying adjustments to check if adjustment date is after the sentencedDate (rather than sentencedDate + 1)

yasinmustafa:


src/test/resources/test_data/overall_calculation/custom-examples/crs-1070-ual-on-tused.json
```diff
@@ -0,0 +1,355 @@
...
+          "MONTHS":0
+        }
+      },
+      "fineAmount":919.72,
+      "identifier":"f9c7ff8a-25ba-39ba-9967-77f94bcdfc16",
+      "recallType":null,
+      "sentencedAt":"2022-04-25",
+      "caseSequence":4,
+      "lineSequence":3,
+      "caseReference":"16008673W CL",
```
ldlharper, line 151: You may want to remove all the data here that isn't required and also isn't anonymous.

ldlharper (changes requested):

src/test/resources/test_data/overall_calculation/custom-examples/crs-1070-ual-on-tused.json
yasinmustafa: Thanks

src/test/resources/test_data/overall_calculation/custom-examples/crs-1070-ual-on-tused.json
```diff
@@ -0,0 +1,355 @@
+{
+  "offender":{
+    "reference":"A8857AD",
```
ldlharper, line 3: DOB and reference also should probably be changed. You can copy the test one from another file.

PR #308, opened 2022-10-28
CRS-1065 Add validation for unsupported adjustments; i.e. LAL and SR

yasinmustafa:
Also did some refactoring as part of this.
- Moved all the validation messages out of the service class and into the `ValidationCode` enum
- Don't wrap api response in `ResponseEntity`
- When the validation succeeds return a object that indicates the success, rather than no-content

TODO determine if we can keep all validation messages in a single place. At the moment, validation is done on the API with messages attached, but then on the UI they are remapped to other messages


joelstobart-moj (approved):
LGTM

PR #309, opened 2022-10-28
CRS-1077 Refactor some data class names so they are not pluralised (in the case where they represent a single object)

yasinmustafa:


joelstobart-moj (approved):
LGTM

PR #313, opened 2022-11-22
CRS-1087 Replace return-to-custody prison-api endpoint call (deprecated) with  fixed-term-recall call

yasinmustafa:
Light touch refactor to replace prison-api call  to return-to-custody-date .

More could be done with this; ideally the ReturnToCustodyDate object should be replaced with FixedTermRecallDetails, however there is quite a lot of change required for this (UI and API). It is being persisted to a database JSON column  and then resurfaced in an endpoint, hence went for the light touch approach for now


ldlharper (approved):
LGTM as a first step. You’d be able to make use of the version columns to save the data to the database and transform the old format.

PR #314, opened 2022-11-23
CRS-1059 updating database structure to allow the saving of ERSED cal…

ldlharper:
…culation user input

### DIFF
--- a/src/main/resources/migration/common/V17__rename_user_input_tables.sql
+++ b/src/main/resources/migration/common/V17__rename_user_input_tables.sql
@@ -26,3 +26,6 @@ UPDATE calculation_request_sentence_user_input sentence_user_input
 SET sentence_user_input.calculation_request_user_input_id = user_input.id
 FROM calculation_request_user_input user_input
 WHERE sentence_user_input.calculation_request_id = sentence_user_input.calculation_request_id;
+
+-- We no longer need the calculation_request_id on the old table. However it won't be dropped until migration is successful.
+ALTER TABLE calculation_request_sentence_user_input ALTER COLUMN calculation_request_id DROP NOT NULL;
--- a/src/main/resources/migration/common/V17__rename_user_input_tables.sql
+++ b/src/main/resources/migration/common/V17__rename_user_input_tables.sql
@@ -13,7 +13,7 @@ CREATE TABLE calculation_request_user_input
 
 
 INSERT INTO calculation_request_user_input(calculation_request_id)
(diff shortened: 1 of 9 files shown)


src/main/resources/migration/common/V17__rename_user_input_tables.sql
```diff
@@ -0,0 +1,28 @@
+ALTER TABLE calculation_request_user_input
```
ldlharper, line 1: Anyone know the best practice for testing a migration script like this? I could take a database dump from dev and test it runs as expected.

matt-crowson-moj: I was wondering if this could be done slightly differently.

Perhaps leave the original table in place and do a `create tables as select ....`

Release it, if it works add another migration for cleaning up the old table and then release that.

Just a thought.

ldlharper: I think the only potential issue is dropping the column. The other changes don't result in any loss of data.

I think I could leave column there, do the release to dev and test it. Then only cleanup the column once confident.

matt-crowson-moj: Sounds reasonable to me.

src/main/resources/migration/common/V17__rename_user_input_tables.sql
```diff
@@ -0,0 +1,31 @@
...
+SELECT DISTINCT calculation_request_id FROM calculation_request_sentence_user_input;
+
+-- Add column to link the sentence user inputs to new table.
+ALTER TABLE calculation_request_sentence_user_input ADD COLUMN calculation_request_user_input_id integer NOT NULL references calculation_request_user_input (id);
+
+-- Set the new column to match data from old table to new up.
+UPDATE calculation_request_sentence_user_input sentence_user_input
+SET sentence_user_input.calculation_request_user_input_id = user_input.id
+FROM calculation_request_user_input user_input
+WHERE sentence_user_input.calculation_request_id = sentence_user_input.calculation_request_id;
```
yasinmustafa, line 28: I may be misunderstanding but this last line doesnt look right, shouldn't
`WHERE sentence_user_input.calculation_request_id = sentence_user_input.calculation_request_id;`
be
`WHERE sentence_user_input.calculation_request_id = user_input.calculation_request_id;`

ldlharper: Yea good spot!!

PR #315, opened 2022-11-24
CRS-1101 Remove validation call from the get-calc-breakdown endpoint, this is unnecessary and causes problems for the view journey

yasinmustafa:
Validation is not required by the view journey as we are just displaying a previous calc. This change also removes the validation call from the get-breakdown call.

Additionally, if we add new validation rules there is a potential for the view journey to break, because it calls validation using old input data (potentially).. and the old input data could break the validation if there are new rules.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BookingService.kt
@@ -18,7 +18,12 @@ class BookingService(
     val prisonerDetails = prisonApiSourceData.prisonerDetails
     val sentenceAndOffences = prisonApiSourceData.sentenceAndOffences
     val bookingAndSentenceAdjustments = prisonApiSourceData.bookingAndSentenceAdjustments
-    validate(prisonApiSourceData)
+    val validationMessages = validationService.validate(prisonApiSourceData)
+    if (validationMessages.isNotEmpty()) {
+      var message = "The validation has failed with errors:"
+      validationMessages.forEach { message += "\n    " + it.message }
+      throw ValidationException(message)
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BookingService.kt
```diff
@@ -34,4 +36,23 @@ class BookingService(
...
+  fun getCalculationBreakdown(
+    calculationRequestId: Long
+  ): CalculationBreakdown {
+    val calculationUserInputs = calculationTransactionalService.findUserInput(calculationRequestId)
+    val prisonerDetails = calculationTransactionalService.findPrisonerDetailsFromCalculation(calculationRequestId)
+    val sentenceAndOffences = calculationTransactionalService.findSentenceAndOffencesFromCalculation(calculationRequestId)
+    val bookingAndSentenceAdjustments = calculationTransactionalService.findBookingAndSentenceAdjustmentsFromCalculation(calculationRequestId)
+    val returnToCustodyDate = calculationTransactionalService.findReturnToCustodyDateFromCalculation(calculationRequestId)
+    val calculation = calculationTransactionalService.findCalculationResults(calculationRequestId)
+    val booking = Booking(
```
ldlharper, line 49: Would it be possible to refactor this so the creation of the Booking is not duplicated? Maybe the validation shouldn't be within the getBooking funciton.

yasinmustafa: Potentially.. I was going to look at that next. Didn't want to do too big a PR in one go.  Will continue the discussion.

ldlharper (approved):
LGTM

PR #325, opened 2022-12-06
CRS-1078 Some rework of FTR validation rules + unit tests + some refactoring of how tests are run + one thousand calc validation

yasinmustafa:
Aside from the validation rules did some extra work:

- Refactored how the tests are run to be inline with the template project (Even int tests are run as unit tests now) (was trying to nest tests and the old setup want allowing me to do so, therefore went down this route)
- The test endpoint (one thousand calcs) now handles validation differently, doesnt throw exceptions
- try catch from full validation is redundant so removed
- Moved the IntTests into the main test package (aligns with the template project and the new way if running tests)
-


.circleci/config.yml
```diff
@@ -26,71 +23,7 @@ jobs:
...
-      tag: "18.0"
-    steps:
-      - checkout
-      - restore_cache:
-          keys:
-            - gradle-{{ checksum "build.gradle.kts" }}
-            - gradle-
-      - run:
-          command: ./gradlew -Dorg.gradle.jvmargs="--illegal-access=permit" -Dkotlin.daemon.jvm.options="--illegal-access=permit" integrationTest
+          command: ./gradlew check
```
yasinmustafa, line 26: brought this inline with the template version, `check` runs all tests (which includes int tests)

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/CalculationResults.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.model
+
+import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.validation.ValidationMessage
+
+data class CalculationResults(
```
yasinmustafa, line 5: This is currently only used by the 1000 calcs, but the plan is to use it for the calulation route as well; i.e. instead of having a separate flow for validation, the validation runs whenever a calculation is requested and messages are returned if it failed.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/CalculationTransactionalService.kt
```diff
@@ -52,24 +51,23 @@ class CalculationTransactionalService(
...
-    try {
-      // TODO Doesn't look like this calculate call actually throws any CrdCalculationValidationException, so this whole try catch block can be removed
-      val bookingAfterCalculation = calculationService.calculate(booking) // Validation stage 2 of 3
-      messages = validationService.validateBookingAfterCalculation(bookingAfterCalculation) // Validation stage 3 of 3
-    } catch (validationException: CrdCalculationValidationException) {
-      return listOf(ValidationMessage(validationException.validation))
-    }
+    messages = validationService.validateBeforeCalculation(booking) // Validation stage 2 of 4
+    if (messages.isNotEmpty()) return messages
+    val bookingAfterCalculation = calculationService.calculate(booking) // Validation stage 3 of 4
```
yasinmustafa, line 69: try catch block removed.. no CrdCalculationValidationException's are being thrown in the calculate call.

build.gradle.kts
```diff
@@ -10,25 +10,6 @@ configurations {
   testImplementation { exclude(group = "org.junit.vintage") }
 }
```
yasinmustafa, line 12: aligning to the template project - tests are run in one stage now and int tests can be run as JUnit tests now

.circleci/config.yml
ldlharper: I think we want the doc linting stuff

joelstobart-moj (approved):
LGTM

ldlharper (dismissed):
LGTM

ldlharper (approved):
LGTM

PR #335, opened 2022-12-19
CRS-1132 Change cache refresh for bank holidays to 8 hours and also add logging to aid diagnosis of any potential timeout issues

yasinmustafa:


joelstobart-moj (approved):
LGTM

PR #340, opened 2023-01-05
Crs 1131 scheduled task to persist bank holiday

ldlharper:


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/repository/BankHolidayCacheRepository.kt
```diff
@@ -0,0 +1,11 @@
+package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.repository
+
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.entity.BankHolidayCache
+
+@Repository
+interface BankHolidayCacheRepository : JpaRepository<BankHolidayCache, Long> {
+
+  fun findFirstByOrderByIdAsc(): BankHolidayCache?
```
yasinmustafa, line 10: will this get the first matching bank holiday when ordered by ascending? In which case shouldn't it order in  Descending?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BankHolidayCacheSchedule.kt
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.scheduling.annotation.Scheduled
+import org.springframework.stereotype.Component
+
+@Component
+class BankHolidayCacheSchedule(
+  val bankHolidayService: BankHolidayService
+) {
+
+  // Every 5 mins from 03:00 till 03:59 Monday to Friday.
+  @Scheduled(cron = "0 */5 3 * * MON-FRI")
```
yasinmustafa, line 14: In a clustered infrastructure where we have multiple pods (i.e. kubernetes) this will result in this schedule firing multiple times simultaneously (per pod)

There are several ways to solve this problem. On Manage Offences we have used SchedLock (See MO for example)

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BankHolidayService.kt
```diff
@@ -0,0 +1,59 @@
...
+class BankHolidayService(
+  private val bankHolidayApiClient: BankHolidayApiClient,
+  private val bankHolidayCacheRepository: BankHolidayCacheRepository,
+  private val objectMapper: ObjectMapper
+) {
+
+  @Transactional
+  fun refreshCache() {
+    var cached = bankHolidayCacheRepository.findFirstByOrderByIdAsc()
+    if (cached == null || cached.cachedAt.toLocalDate().isBefore(LocalDate.now())) {
```
yasinmustafa, line 24: Not sure what the purpose of this if condition is?

ldlharper: So the idea is that the schedule will run every 5mins for an hour in the evening. The re-running is important in case the request fails. 

This condition checks if the cache is empty, or if the cache has not yet been updated today.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BankHolidayCacheSchedule.kt
ldlharper: Yea, so the code was written so that if the scheduled job was picked up by two different services at the same time, the logic would still work and update the cache.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/repository/BankHolidayCacheRepository.kt
ldlharper: So there is only ever one entry in the database. The schedule will just update the row every night as there is no need to capture historic bank holiday data.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BankHolidayCacheSchedule.kt
yasinmustafa: If multiple pods run simultaneously then isn't there a chance that they will both insert - because the commit isn't done until the transaction is finished? To discuss

ldlharper: Yea that could happen. But only on the first ever run of the schedule. Subsequent runs will be updates, and the updates will just update the first row and the second row would be ignored forever.

I had a quick look and maybe a simple thing to do would make the second the job first on random.

https://stackoverflow.com/a/41018664

  @Scheduled(cron = "${random.int[0,59]} */5 3 * * MON-FRI")

yasinmustafa: sounds good to me

joelstobart-moj (approved):
LGTM

PR #344, opened 2023-01-09
CRS-1153 removing spring cache from bank holiday.

ldlharper:


matt-crowson-moj:
@joelstobart-moj can I ask why you are removing the caching?  We use this feature in our activities service for bank holidays.

yasinmustafa (approved):
LGTM

PR #356, opened 2023-02-08
CRS-1184 fix for DTOs

jbowdidge-moj:


joelstobart-moj (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/TransformFunctions.kt
```diff
@@ -147,54 +147,54 @@ fun transform(sentence: SentenceAndOffences, calculationUserInputs: CalculationU
       )
     } else {
       val imprisonmentTerm = sentence.terms.first { it.code == SentenceTerms.IMPRISONMENT_TERM_CODE }
-      val licenseTerm = sentence.terms.first { it.code == SentenceTerms.LICENCE_TERM_CODE }
```
ldlharper, line 150: The reason for this separation of code is that SOPC + EDS have two different terms (license and imprisonment). I think the DTO condition lives up above as another `else if` statement, alongside Afine and SDS

ldlharper (approved):
LGTM

PR #365, opened 2023-02-24
CRS-1219 Change SLED to SED for DTOs

jbowdidge-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BookingExtractionService.kt
```diff
@@ -458,9 +458,15 @@ class BookingExtractionService(
...
     val hasLicence = sentences.any() { it.sentenceCalculation.licenceExpiryDate != null && it.sentenceCalculation.licenceExpiryDate!!.isAfterOrEqualTo(latestReleaseDate) }
 
-    if ((sentences.any { it.isDto() } && booking.underEighteenAtEndOfCustodialPeriod())) {
-      return ConcurrentOraAndNonOraDetails(isReleaseDateConditional = false, canHaveLicenseExpiry = true)
+    if ((sentences.any { it.isDto() })) {
+      if (booking.underEighteenAtEndOfCustodialPeriod()) {
+        return ConcurrentOraAndNonOraDetails(isReleaseDateConditional = false, canHaveLicenseExpiry = true)
+      }
+      if (sentences.any { !it.isDto() } && !booking.underEighteenAtEndOfCustodialPeriod()) {
+        return ConcurrentOraAndNonOraDetails(isReleaseDateConditional = true, canHaveLicenseExpiry = true)
```
ldlharper, line 466: Is this change related? The only thing that changes here is that the release date is now always conditional if the booking contains DTO and non DTO sentences.

ldlharper (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BookingExtractionService.kt
jbowdidge-moj: As per convo with Cherky, this is what I believe to be true. Happy to make adjustments if we discover this isn't quite right.

PR #399, opened 2023-04-05
MDWFP-73 adding relevant remand endpoint.

ldlharper:
Adding endpoint to be used by the relevant remand tool.


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/CalculationTransactionalService.kt
```diff
@@ -84,6 +90,26 @@ class CalculationTransactionalService(
     }
   }
 
+  fun relevantRemandCalculation(prisonerId: String, request: RelevantRemandCalculationRequest): RelevantRemandCalculationResult {
```
jbowdidge-moj, line 93: Is the point of this class, that it deals with storing data in the database? Should this function live here?

jbowdidge-moj (approved):
LGTM

PR #402, opened 2023-04-12
CRS-1325 Enable WAF (modsecurity) in detection only mode

yasinmustafa:


helm_deploy/calculate-release-dates-api/values.yaml
```diff
@@ -13,8 +13,6 @@ generic-service:
     enabled: true
     host: app-hostname.local    # override per environment
     tlsSecretName: calculate-release-dates-api-cert
-    v1_2_enabled: true
```
yasinmustafa, line 16: these flags aren't required any more

joelstobart-moj (approved):
LGTM

PR #408, opened 2023-05-03
MDWFP-114 fix an issue with the relevant remand calculation.

ldlharper:
Fixing issue where there is too much relevant remand


joelstobart-moj (approved):
LGTM

PR #431, opened 2023-06-26
A few changes to the API

joelstobart-moj:
- Change the endpoint value
- Number of people compared
- Update test to match new API
- Add Dummy


jbowdidge-moj (approved):
LGTM

PR #439, opened 2023-07-19
CRS-1496 Store approved dates alongside a calculation 

jbowdidge-moj:


ldlharper (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/entity/ApprovedDates.kt
```diff
@@ -0,0 +1,27 @@
...
+import jakarta.validation.constraints.NotNull
+import java.time.LocalDate
+
+@Entity
+@Table
+data class ApprovedDates(
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  val id: Long = -1,
+  @NotNull
```
yasinmustafa, line 19: Question about the @NotNull annotation - does it provide anything additional to the Kotlin default behaviour; i.e. the absence of a ? implies not null

src/main/resources/migration/common/V24__approved_dates.sql
```diff
@@ -0,0 +1,22 @@
+create table approved_dates_submission
+(
+    id                     serial
+        constraint approved_dates_request_pk primary key,
+    calculation_request_id integer                  NOT NULL references calculation_request (id),
+    prisoner_id            varchar(10)              not null,
```
yasinmustafa, line 6: aren't some of the fields here below duplicated in the calculation request table? e.g prisonerId, booking Id?

could we piggy back from those values? or maybe faster to read if we stamp here as well

jbowdidge-moj: They are duplicated yes. There will be instances when the approved dates for a prisoner could be 'removed', so having them also stored here will make that easier. It's a small amount of future proofing. There will also be instances where we want to look up just the approved dates without getting all of the calculation information (e.g on hub) so having it here will make that easier.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/CalculationTransactionalService.kt
```diff
@@ -118,11 +127,15 @@ class CalculationTransactionalService(
     sourceData: PrisonApiSourceData,
     booking: Booking,
     userInput: CalculationUserInputs?,
+    approvedDates: List<ManualEntrySelectedDate>?,
   ): CalculatedReleaseDates {
     try {
       val calculation =
         calculate(booking, CONFIRMED, sourceData, userInput, calculationFragments)
-      writeToNomisAndPublishEvent(prisonerId, booking, calculation)
+      if (approvedDates != null) {
```
yasinmustafa, line 135: just wondering if there is any benefit in splitting out the 'approved dates storage' into a separate endpoint? 

Although looks like it uses the calculation so makes sense to leave it here

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/entity/ApprovedDates.kt
jbowdidge-moj: I think this needs more investigation!

PR #446, opened 2023-08-11
CRS-1536 store the living unit details with the prisoner details at t…

jbowdidge-moj:
…he time of calculation


yasinmustafa (approved):
LGTM

PR #458, opened 2023-09-29
CRS-1640 link calcs together when Specialist support agrees with CRDS

jbowdidge-moj:


joelstobart-moj (approved):
LGTM - I'm not a huge fan of the try catch. Would be good to find a way to make that less ugly :(

PR #474, opened 2023-10-20
CRS-1667 use max in memory size config

carloveo-moj:


src/main/resources/application.yml
```diff
@@ -6,7 +6,7 @@ spring:
   application:
     name: calculate-release-dates-api
   codec:
-    max-in-memory-size: 10MB
+    max-in-memory-size: 200MB
```
carloveo-moj, line 9: this doesn't need to be this drastic, the response in dev was a little over 10MB

joelstobart-moj (approved):
LGTM

PR #485, opened 2023-11-01
Adjust1 248 unused deductions calculation

ldlharper:
This PR creates an endpoint that will re-calculate the unused deductions and save the data in adjustments api.

This was going to be event based, but the auth and sqs setup was taking a while and probably too much for a PoC. If we go with the event approach it will also probably be handled in a different service.


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/unuseddeductions/service/UnusedDeductionsCalculationService.kt
```diff
@@ -0,0 +1,74 @@
...
+    val releaseDateTypes = listOf(ReleaseDateType.CRD, ReleaseDateType.ARD, ReleaseDateType.MTD)
+    val calculationResult = result.second
+    var releaseDate = calculationResult.dates.filter { releaseDateTypes.contains(it.key) }.minOfOrNull { it.value }
+    val releaseDateSentence = result.first.getAllExtractableSentences().find { it.sentenceCalculation.releaseDate == releaseDate }
+    val unusedDeductions = if (releaseDateSentence != null) { // The release date will be a PRRD, no unused deductions.
+      val maxDeductions =
+        releaseDateSentence.sentenceCalculation.numberOfDaysToDeterminateReleaseDate
+
+      val remand = adjustments.filter { it.adjustmentType == AdjustmentServiceAdjustmentType.REMAND }.map { it.daysBetween!! }.reduceOrNull { acc, it -> acc + it } ?: 0
+      val taggedBail = adjustments.filter { it.adjustmentType == AdjustmentServiceAdjustmentType.TAGGED_BAIL }.map { it.days!! }.reduceOrNull { acc, it -> acc + it } ?: 0
```
iterative-it, line 51: Are daysBetween and days guaranteed to not be null here? Would it make sense to filter out any null values first?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/unuseddeductions/service/UnusedDeductionsCalculationService.kt
```diff
@@ -0,0 +1,74 @@
...
+      AdjustmentServiceAdjustmentType.RESTORATION_OF_ADDITIONAL_DAYS_AWARDED -> BookingAdjustmentType.RESTORED_ADDITIONAL_DAYS_AWARDED
+      AdjustmentServiceAdjustmentType.ADDITIONAL_DAYS_AWARDED -> BookingAdjustmentType.ADDITIONAL_DAYS_AWARDED
+      else -> null
+    }
+  }
+
+  private fun useAdjustmentsFromAdjustmentsApi(adjustments: List<AdjustmentServiceAdjustment>): BookingAndSentenceAdjustments {
+    return BookingAndSentenceAdjustments(
+      bookingAdjustments = adjustments.filter { mapToBookingAdjustmentType(it.adjustmentType) != null }
+        .map { BookingAdjustment(active = true, fromDate = it.fromDate!!, toDate = it.toDate, numberOfDays = it.effectiveDays!!, type = mapToBookingAdjustmentType(it.adjustmentType)!!) },
```
iterative-it, line 70: Similar question to above here. If in reality the API always populates these properties, you could consider making them non nullable in the data class. Appreciate I don't know this API or the context as yet! :)

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/unuseddeductions/service/UnusedDeductionsCalculationService.kt
ldlharper: Yes they've been filtered out in the previous class. They're only null if the record was created and synchronised from NOMIS.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/unuseddeductions/service/UnusedDeductionsCalculationService.kt
ldlharper: Yea I think we may want to improve this. Currently most things in the DTO are nullable because the record could've been created in NOMIS where none of these fields are required. We could maybe have two different DTOs depending on the source of the record.

iterative-it (approved):
LGTM but you might want to wait for another approval as I'm new to this project :)

ldlharper:
> LGTM but you might want to wait for another approval as I'm new to this project :)

Thank you :)

PR #493, opened 2023-11-10
ADJUST1-276 Unused deductions infrastructure.

ldlharper:
This PR adds an endpoint to the unused deductions that will only calculate the unused deductions. The updating of adjustments functionality will be removed from calculate release dates and be handled by the new unused deductions listener where the changes are handled with sqs queues instead.


joelstobart-moj (approved):
LGTM

PR #495, opened 2023-11-15
CRS-1693: Populate prisoner_location from  prisoner_details where available.

justin-gardiner:


jbowdidge-moj (approved):
LGTM - Tested the queries as a select against dev.

PR #509, opened 2023-11-28
CRS-1696 Backend | Add a new reference data table.

dave-virden:
* Creates and populates a table to hold the calculation reasons.
* Links the calculation reasons to the calculation request.
* Initial values are loaded from the prototype.
* Existing calculation requests are updated to use the default value.


src/main/resources/migration/common/V37__add_calculation_reason.sql
```diff
@@ -0,0 +1,32 @@
+CREATE TABLE calculation_reason
+(
+    id           INTEGER PRIMARY KEY,
```
jbowdidge-moj, line 3: This should be a generated value.

src/main/resources/migration/common/V37__add_calculation_reason.sql
```diff
@@ -0,0 +1,32 @@
...
+    is_other     BOOLEAN DEFAULT FALSE
+);
+
+ALTER TABLE calculation_request
+    ADD COLUMN reason_for_calculation INT REFERENCES calculation_reason (id);
+ALTER TABLE calculation_request
+    ADD COLUMN other_reason_for_calculation VARCHAR(40);
+
+UPDATE calculation_request
+SET reason_for_calculation = -1
```
jbowdidge-moj, line 15: You can't do the update until you've inserted the data because of the foreign key constraint.

src/main/resources/migration/common/V37__add_calculation_reason.sql
```diff
@@ -0,0 +1,32 @@
...
+WHERE reason_for_calculation IS NULL;
+
+INSERT INTO calculation_reason (id, rank, display_name)
```
jbowdidge-moj, line 18: I'm not sure why we're ranking them? Shouldn't they just be `active` or `inactive`?

src/main/resources/migration/common/V37__add_calculation_reason.sql
jbowdidge-moj, line 15: Because you don't know the ID because it'll be auto generated you'll need to run an inner select to get the appropriate ID.

src/main/resources/migration/common/V37__add_calculation_reason.sql
dave-virden: My thought was a rank so the order can be preserved in the front end, there may be another method of doing this?

jbowdidge-moj (approved):
LGTM

PR #510, opened 2023-11-30
CRS-1709 Use new paginated prison-api end point to replace existing unpaged version (for retrieving all Calculable Sentences for bookings in a establishment)

yasinmustafa:


joelstobart-moj (approved):
LGTM

PR #513, opened 2023-12-01
CRS-1748 Backend | Create endpoint to supply reasons for the frontend

dave-virden:
* Create and test an endpoint to return active calculation reasons.


jbowdidge-moj (approved):
LGTM

PR #526, opened 2023-12-18
CRS-1769: Classify unsupported calculation types as unsupported sente…

justin-gardiner:
…nce types in bulk comparisons.


iterative-it (approved):
LGTM

PR #536, opened 2024-01-04
CRS-1746 add HDCED4PLUS to bulk comparison results

jbowdidge-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BulkComparisonService.kt
```diff
@@ -109,6 +114,22 @@ class BulkComparisonService(
     comparisonRepository.save(comparison)
   }
 
+  private fun getHdced4PlusDate(mismatch: Mismatch): LocalDate? {
+    val hdced4Plus = mismatch.calculatedReleaseDates?.dates?.getOrDefault(ReleaseDateType.HDCED4PLUS, null)
+    val hdced = mismatch.calculatedReleaseDates?.dates?.getOrDefault(ReleaseDateType.HDCED, null)
+    if (hdced4Plus == null && hdced == null) {
```
carloveo-moj, line 120: I'm not sure you need the two if statements before the return? I think the return if statement covers these two scenarios?

carloveo-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BulkComparisonService.kt
jbowdidge-moj: I think this expresses the business rules clearer...rather than implying them. I'll keep it unless it becomes a problem.

PR #537, opened 2024-01-05
CRS-1778: Add a new mismatch type for unsupported sentences that coul…

justin-gardiner:
…d be potentially be eligible for HDC4+.


dave-virden (approved):
LGTM

PR #540, opened 2024-01-11
CRS-1779: Update NOMIS key dates with the reason.

dave-virden:
* Corresponding changes to the prison api have already been made https://github.com/ministryofjustice/prison-api/pull/1975


jbowdidge-moj (changes requested):
This definitely meets the necessary criteria, but I think we could do some refactoring on the historic code to make this neater. I'm thinking probably extract all the comments to a `NomisCommentService` that can just return the comment would make this a lot more readable. It would also put all the comments in a central place if we ever needed to change them in the future. Happy to discuss.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/NomisCommentService.kt
```diff
@@ -0,0 +1,65 @@
...
+  fun getNomisComment(
+    calculationRequest: CalculationRequest,
+    isSpecialistSupport: Boolean,
+    approvedDates: List<ManualEntrySelectedDate>?,
+  ): String {
+    val comment = if (isSpecialistSupport) {
+      SPECIALIST_SUPPORT_COMMENT
+    } else if (approvedDates?.isNotEmpty() == true) {
+      MANUAL_ENTERED_DATES_COMMENT
+    } else if (!calculationRequest.reasonForCalculation?.isOther!!) {
```
jbowdidge-moj, line 21: I think this will handle the null case a bit better:

`calculationRequest.reasonForCalculation?.isOther == false`

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/NomisCommentService.kt
```diff
@@ -0,0 +1,65 @@
...
+      )
+    }
+  }
+
+  fun getManualNomisComment(
+    calculationRequest: CalculationRequest,
+    dates: Map<ReleaseDateType, LocalDate?>,
+    isGenuineOverride: Boolean,
+  ): String {
+    var comment = if (dates.containsKey(ReleaseDateType.None)) INDETERMINATE_COMMENT else MANUAL_ENTRY_COMMENT
```
jbowdidge-moj, line 42: I think this is unnecessary reassignment, this could be:
```
var comment = if (isGenuineOverride) MANUALLY_ENTERED_OVERRIDE 
else if (dates.containsKey(ReleaseDateType.None)) INDETERMINATE_COMMENT 
else MANUAL_ENTRY_COMMENT
```

jbowdidge-moj (dismissed):
Just two small tweaks on this one :)

srowlands-moj (dismissed):
LGTM  - it was fun piecing together the reasons that already existed from previous incrementals vs the list in the ticket though 🙃

dave-virden:
Please could I have another review on this, I've just rebased it on to the top main

srowlands-moj (approved):
LGTM

PR #546, opened 2024-01-19
CRS-875 Display reason for calculation in the view journey

dave-virden:
* Added the reason the to CalculatedReleaseDates so that it can be displayed in the view journey.
* Updated the tests to reflect this change.


joelstobart-moj:
HI @dave-virden  - Could you add a screenshot or a video to this PR? It would be helpful to get an idea of the impact.

joelstobart-moj (approved):
LGTM

Can you make sure that the shape of this aligns between the UI and API ?

PR #547, opened 2024-01-22
CRS-1750: Add new db tables to store bulk comparison discrepancy data.

justin-gardiner:


src/main/resources/migration/common/V44__add_comparison_discrepancy_tables.sql
```diff
@@ -0,0 +1,46 @@
...
+
+create table comparison_person_discrepancy
+(
+    id serial constraint comparison_person_discrepancy_pk primary key,
+    comparison_person_id integer not null,
+    impact_id integer not null,
+    priority_id integer not null,
+    detail text,
+    action text,
+    created_by varchar(20) not null,
```
srowlands-moj, line 21: I just raised a ticket https://dsdmoj.atlassian.net/browse/CRS-1794 as I noticed that fields capturing username are a bit all over the place in their length.

justin-gardiner: ok, I've updated to 40 for now but can update again when someone has a look at that ticket

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/entity/ComparisonPersonDiscrepancyCategory.kt
```diff
@@ -0,0 +1,36 @@
...
+
+  @NotNull
+  @Enumerated(EnumType.STRING)
+  val category: DiscrepancyCategory,
+
+  @NotNull
+  @Enumerated(EnumType.STRING)
+  val subCategory: DiscrepancySubCategory,
+
+  val other: String? = null,
```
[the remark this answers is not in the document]
justin-gardiner: yes, it does look a bit out of place but it is needed. The user can enter one or more causes for the discrepancy where each cause has a category, subcategory and an optional other field where they can enter more detail. I've renamed the class and field to hopefully make it a bit clearer. This is only the first sub task on the story so may get updated some more when I start the other tasks.

srowlands-moj (approved):
LGTM

PR #557, opened 2024-02-02
Crs 1785: Make the typescript generation work

dave-virden:


srowlands-moj (approved):
LGTM

srowlands-moj:
I'd be interested to know what issues this caused the type generation though

dave-virden:
> I'd be interested to know what issues this caused the type generation though

It was being cast to ValidationMessage in the front end, so every time you ran the script to generate the type in the front end it had to be manually changed from JsonNode

PR #566, opened 2024-02-05
CRS-1788 Fix issue with not all HDC4+ not being shown in the list

srowlands-moj:
* Some calculated HDCED4+ results were not being returned due to the rep query, they are now returned in a seperate result list.


jbowdidge-moj (approved):
LGTM

PR #569, opened 2024-02-08
Revert "Crs 1785: Make the typescript generation work"

dave-virden:
Reverts ministryofjustice/calculate-release-dates-api#557

Changes to the db need to be considered


srowlands-moj:
Can we not just change the persistence type back to what it was but change the DTO to return a converted list

dave-virden:
Yes, that makes sense

PR #571, opened 2024-02-09
ComparisonMismatchSummary now returns ValidationMessages

dave-virden:
* Transformation is now carried out on the DTO object

### DIFF
--- a/run-local.sh
+++ b/run-local.sh
@@ -24,6 +24,9 @@ export DB_PASS=calculate-release-dates
 # export HMPPS_AUTH_URL=https://sign-in-dev.hmpps.service.justice.gov.uk/auth
 export HMPPS_AUTH_URL=http://localhost:9090/auth
 
+# export PRISON_API_URL=https://prison-api-dev.prison.service.justice.gov.uk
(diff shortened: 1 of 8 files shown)


james-baker-digital-justice:
I would suggest a test that checks it serialises some ValidationMessage as well as empty list.

run-local.sh
```diff
@@ -24,6 +24,9 @@ export DB_PASS=calculate-release-dates
 # export HMPPS_AUTH_URL=https://sign-in-dev.hmpps.service.justice.gov.uk/auth
 export HMPPS_AUTH_URL=http://localhost:9090/auth
 
+# export PRISON_API_URL=https://prison-api-dev.prison.service.justice.gov.uk
+export PRISON_API_URL=http://localhost:8080
+
```
james-baker-digital-justice, line 29: This seems unrelated to this change, just checking if you meant to include it. (otherwise LGTM)

dave-virden: Yes, I'd fallen over it a couple of times and left it in as it makes it easier to setup the local environment if you're not familiar with it.

james-baker-digital-justice (approved):

PR #572, opened 2024-02-14
CRS-1811 - Add release date to HDC4+ mismatches in bulk comparison

james-baker-digital-justice:
Split the model for HDC4+ and other comparison tabs as they had a separate set of entities required for each. Did some minor refactoring to allow overrides of the prison API wiremock. This allowed setup of more complicated tests without impacting other tests.


justin-gardiner (approved):
LGTM

PR #573, opened 2024-02-14
Crs 1814 consecutive sentences for hdc4 unsupported

justin-gardiner:


james-baker-digital-justice (approved):
Looks good to me but I have limited understanding of the logic so take with a pinch of salt

PR #574, opened 2024-02-15
CRS-1811: match up valuation outcomes with people on HDC4+ mismatches

james-baker-digital-justice:
Was taking the latest across any person instead of valuation outcomes per person.


carloveo-moj (approved):
LGTM

PR #581, opened 2024-02-19
CRS-1814: Replace calculation outcome repository method findByCalcula…

justin-gardiner:
…tionRequestIdIn. When running large comparisons this was being passed a list of thousands of id's and was getting very slow.


srowlands-moj (approved):
LGTM

PR #582, opened 2024-02-20
CRS-1827: Fix all prisons comparison number of comparisons and status…

justin-gardiner:
… being updated to complete after one prison processed.


james-baker-digital-justice (commented):
should there be a test that manual comparison updates state to complete as well?

justin-gardiner:
@james-baker-digital-justice Yes there should be a test for it, I assumed there was an existing test but it seems there wasn't. I've added one.

james-baker-digital-justice (approved):
Thanks. My spidey sense was tingling when I saw 3 changes in src/main and 2 in src/test :)

PR #585, opened 2024-02-27
CRS-1634 Add verification for manual entry submission

srowlands-moj:
* Add new controller endpoint for manual entry verification (Currently only validates the offence dates so that it can be successfully submitted back to NOMIS)
* Made the validation more understandable
* Bump the Jacoco plugin version to remove the sea of red when running tests


james-baker-digital-justice (approved):
LGTM

PR #586, opened 2024-02-28
Crs 1831 Update "Reason for Calculation" Options

dave-virden:


srowlands-moj (approved):
LGTM - only question is are we intending on having the flexibility of maintaining displayname and nomis comment

dave-virden:
> LGTM - only question is are we intending on having the flexibility of maintaining displayname and nomis comment

It's a least worst option, that gives the users a better experience and doesn't break Nomis without resulting to arbitrary substringing, it shouldn't change too much once deployed

PR #589, opened 2024-03-07
Crs 1541: Regression HDCED unit test scenarios

dave-virden:
Four additional tests have been added to shore up the calculation engine with a bit of refactoring.


james-baker-digital-justice (approved):
LGTM but I don't really understand the problem so up to you if you want someone else to check before you merge

PR #594, opened 2024-03-15
CRS-1848: Show that prisoner detail and offences are historic

james-baker-digital-justice:
Minor model change to show that prisoner detail and sentence and offence information is that which was used to make the calculation and is not current.


justin-gardiner (approved):
LGTM

PR #595, opened 2024-03-15
CRS-1847: Reference data API

james-baker-digital-justice:
API to supply the release date types and their descriptions to the front end.


dave-virden (approved):
LGTM

PR #598, opened 2024-03-19
CRS-1730: In comparison discrepancy endpoints check establishment on …

justin-gardiner:
…comparison person instead of comparison as for a full prison comparison the comparison establishment is "all".


joelstobart-moj (approved):
LGTM

PR #601, opened 2024-03-20
CRS-1849: Latest calc model changes

james-baker-digital-justice:
Remove use of map and just use a list as this is to present all dates rather than pick specific dates. This preserves the order of dates as well. Make calculatedAt mandatory in line with latest prison API.


joelstobart-moj (approved):
LGTM

PR #602, opened 2024-03-20
CRS-1849: Add booking id to latest calc API

james-baker-digital-justice:
This is required to be able to link into the view journey which requires prisoner id and booking id.


srowlands-moj (approved):
LGTM

PR #604, opened 2024-03-20
CRS-1849: Add calc request id to latest calc API

james-baker-digital-justice:
This is required for linking latest calc component up with other CRDS journeys and not calculation reference that I originally thought...


justin-gardiner (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/resource/CalculationControllerTest.kt
```diff
@@ -365,7 +365,7 @@ class CalculationControllerTest {
       prisonerId,
       123456,
       LocalDateTime.now(),
-      UUID.randomUUID().toString(),
+      654321,
```
joelstobart-moj, line 368: Should this be a UUID string?

james-baker-digital-justice: was replacing UUID calc reference with Long calculation request id as I added the wrong one to the API

PR #606, opened 2024-03-20
Bump Prod pod mem limits

srowlands-moj:
* Apply pod memory limit bump that was made on preprod to prevent bulk calc pods from being culled when they reached the 1GB limit.


justin-gardiner (approved):
LGTM

PR #607, opened 2024-03-21
CRS-1849: hide SED and LED if we can create a SLED

james-baker-digital-justice:
Was showing SLED, LED and SED all at once. Only show the SED and LED if they are not the same or are missing. If they are the same only show SLED.


justin-gardiner (approved):
LGTM

PR #616, opened 2024-04-05
CRS-1793: Calculate ERSED using min/max instead of halfway/two thirds

james-baker-digital-justice:
Now calculates at either halfway or -544 days and takes the latter of the two dates. Previously this was only used for concurrent sentences but is now used for single sentences as well simplifying the calculation. Incidentally fixes the hint text for max calc on concurrent sentences.


joelstobart-moj (approved):
LGTM

PR #618, opened 2024-04-08
CRS-1644 Remove CalculationQuestions

srowlands-moj:
* Update OffenceSDSPlusLookupService to set offence flags
* Remove code for calculation questions

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/OffenceSdsPlusLookupService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/OffenceSdsPlusLookupService.kt
@@ -58,7 +58,7 @@ class OffenceSdsPlusLookupService(
                 val sentenceCalculationType = SentenceCalculationType.from(sentenceAndOffence.sentenceCalculationType)
                 val sevenYearsOrMore = sevenYearsOrMore(sentenceAndOffence)
 
-                return@filter isOffenceSdsPlus(
+                checkIsSDSPlusAndSetOffenceIndicators(
                   sentenceCalculationType,
                   sentenceAndOffence,
                   sevenYearsOrMore,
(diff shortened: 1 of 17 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/OffenceSdsPlusLookupService.kt
```diff
@@ -58,7 +58,14 @@ class OffenceSdsPlusLookupService(
                 val sentenceCalculationType = SentenceCalculationType.from(sentenceAndOffence.sentenceCalculationType)
                 val sevenYearsOrMore = sevenYearsOrMore(sentenceAndOffence)
 
-                return@filter isOffenceSdsPlus(sentenceCalculationType, sentenceAndOffence, sevenYearsOrMore, moResponseForOffence, sentenceIsAfterPcsc)
+                return@filter isOffenceSdsPlus(
```
james-baker-digital-justice, line 61: Maybe just rename this to show there are side effects, e.g., checkIsSDSPlusAndSetIndicators. Also` return @filter` is superfluous as it's already a boolean and the last line

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/PrisonService.kt
```diff
@@ -67,8 +68,12 @@ class PrisonService(
   }
 
   fun getSentencesAndOffences(bookingId: Long, filterActive: Boolean = true): List<SentenceAndOffences> {
-    return prisonApiClient.getSentencesAndOffences(bookingId)
+    val sentencesAndOffences = prisonApiClient.getSentencesAndOffences(bookingId)
       .filter { !filterActive || it.sentenceStatus == "A" }
+
+    offenceSdsPlusLookupService.populateSdsPlusMarkerForOffences(sentencesAndOffences)
```
james-baker-digital-justice, line 74: Can this blow up? I see getSentencesAndOffences is used in a number of places.

srowlands-moj: Now throws a sensible error if it does blow up which will get propagated, it'll appear as any normal error does though to the user as does everything for the time being.

PR #620, opened 2024-04-09
CRS-1876: Move release points to configuration

james-baker-digital-justice:
Moved the release points into configuration based on sentence identification track. Also updated all the config to be easily loaded into tests that rely on it rather than duplicating into the tests by hand.


srowlands-moj (approved):
Nice, LGTM

PR #621, opened 2024-04-11
CRS-1882: HDCED hint should use next working day instead of previous

james-baker-digital-justice:
When a HDCED falls on a weekend or bank holiday the weekend adjustment hint text should show the date of the next working day. Other date types remain on previous working day.


dave-virden (approved):
LGTM

PR #632, opened 2024-04-15
CRS-1888: Allow override of calculation params in tests for SDS45

james-baker-digital-justice:
Can be set to calculation-params or alt-calculation-params and will use application-calculation-params.yml or application-alt-calculation-params.yml to source release point config accordingly.


joelstobart-moj (approved):
LGTM

PR #641, opened 2024-04-19
CRS-1938: store version in calc request

james-baker-digital-justice:
One day we're going to need to figure out why we calculated those dates and it'll be hard to figure out the version of the code (which also includes the config). This allows us to do that.


joelstobart-moj (approved):
LGTM

PR #642, opened 2024-04-19
Tests for Alternative Release Point 2

joelstobart-moj:
1) changing to references to "alternative-release-point"
2) Adding test cases for alternative release point 2 
3) Wiring tests for alternative release point 2 which pass (same tests as passing at alt-1 which is positive)


james-baker-digital-justice (approved):
LGTM

PR #644, opened 2024-04-19
CRS-1794: make all username fields 240 chars

james-baker-digital-justice:
Username can by up to 240 according to hmpss-auth schema. In reality, it will probably not be this big as it should only be a staff id for NOMIS users or a UUID for systems however custom users can be created such as CRD_TEST_USER which breaks the usual pattern.


srowlands-moj (approved):
LGTM

PR #647, opened 2024-04-23
Allow ERSED to be recalced post timeline walk

srowlands-moj:
* ERSED was previously not being recalculated post group adjustments taking effect
* Add alt scenario 9 to test list


joelstobart-moj (approved):
LGTM

PR #662, opened 2024-04-30
Adding test case 30 and 32 and including 32 in the test suite

joelstobart-moj:


dave-virden (approved):
LGTM, I assume you're not running 32 on purpose?

PR #666, opened 2024-05-01
CRS-1940: Fix issue with bulk comparison

james-baker-digital-justice:
SDS+ matches in bulk comparison were failing to parse after the SDS+ refactor


srowlands-moj (approved):
LGTM

PR #668, opened 2024-05-03
CRS-1955: Remove is schedule fifteen flag

james-baker-digital-justice:
The flag was never actually set and the sentences are unsupported. Removed the logic and associated tests.


joelstobart-moj (approved):
LGTM

PR #669, opened 2024-05-07
CRS-1942: Fix SDS early release exclusion markers

james-baker-digital-justice:
Wrong value used for no match from manage offences. Should be NONE but was NO


dave-virden (approved):
LGTM

PR #671, opened 2024-05-08
CRS-1942: Refactor before & after CJA & LASPO for new tracks

james-baker-digital-justice:
Removing the before/after CJA & LASPO logic from SentenceIdentificationTrack and moving to the sentence itself.  Adding new tracks for SDS standard, plus and early release points.


joelstobart-moj (approved):
LGTM

PR #672, opened 2024-05-08
Adding in a new endpoint to retrieve just the release dates based on …

FaizJD030:
Introduced a new endpoint /calculation/release-dates/${calcRequestId} to fetch the OffenderReleaseDates based on a calculationRequestId.  

This endpoint enhances the feature of showing the dates on the print notification slip in the frontend. Refer to https://dsdmoj.atlassian.net/browse/CRS-1856 for more details.  

The endpoint /calculation/detailed-results/{calculationRequestId} was evaluated but not chosen due to its extensive functionality beyond just retrieving the OffenderReleaseDates, making it less efficient for this specific use case.


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/OffenderKeyDatesService.kt
```diff
@@ -5,18 +5,40 @@ import org.springframework.stereotype.Service
...
+      val calculationRequest = calculationRequestRepository.findById(calculationRequestId).get()
+      val offenderKeyDatesEither = prisonService.getOffenderKeyDates(calculationRequest.bookingId)
+      val releaseDatesForBookingId = offenderKeyDatesEither.map { releaseDates(it) }.getOrNull()
+      return OffenderReleaseDates(
+        calculationRequest.bookingId,
+        calculationRequestId,
+        calculationRequest.calculatedAt,
+        calculationRequest.reasonForCalculation?.displayName ?: "Not entered",
+        CalculationSource.CRDS,
+        calculationResultEnrichmentService.addDetailToCalculationDates(releaseDatesForBookingId!!, null, null).values.toList(),
```
james-baker-digital-justice, line 35: this is an unsafe null check, would be better to handle the error directly and throw a relevant exception. What does detailed calculation results do in this case? 
You could also make use of the either like
`offenderKeyDatesEither
       .map { releaseDates(it) }
       .map { calculationResultEnrichmentService.addDetailToCalculationDates(releaseDatesForBookingId!!, null, null).values.toList() }
      .fold(
{error -> throw SomeException(error)},
{enrichedDates -> OffenderReleaseDates(...snip...)}
)`

src/test/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/resource/CalculationControllerTest.kt
```diff
@@ -464,5 +465,63 @@ class CalculationControllerTest {
...
+      .isEqualTo(
+        ErrorResponse(
+          HttpStatus.NOT_FOUND,
+          null,
+          errorMessage,
+          errorMessage,
+        ),
+      )
+  }
+
```
james-baker-digital-justice, line 525: missing test for offender key dates returned an error (Left)

src/test/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/OffenderKeyDatesServiceTest.kt
```diff
@@ -74,6 +85,74 @@ open class OffenderKeyDatesServiceTest {
...
+  @Test
+  fun `Test getting Release Dates for calc request id for exception scenario`() {
+    val calcRequestId = 5636121L
+    val errorMessage = "Unable to retrieve offender key dates"
+
+    val exception = assertThrows<CrdWebException> {
+      underTest.getKeyDatesByCalcId(calcRequestId)
+    }
+
+    Assertions.assertThat(exception.message).isEqualTo(errorMessage)
```
james-baker-digital-justice, line 153: add static import?

james-baker-digital-justice (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/OffenderKeyDatesService.kt
FaizJD030: Changed as Requested

src/test/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/resource/CalculationControllerTest.kt
FaizJD030: Add 3 new cases as discussed in OffenderKeyDatesServiceTest.kt

src/test/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/OffenderKeyDatesServiceTest.kt
FaizJD030: Import addressed to be static

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/OffenderKeyDatesService.kt
```diff
@@ -3,20 +3,58 @@ import arrow.core.getOrElse
...
+                calculationRequest.otherReasonForCalculation,
+                calculationRequest.calculatedAt.toLocalDate(),
+                calculationRequest.calculationType,
+              ),
+              enrichedDates,
+            )
+          },
+        )
+    } catch (e: Exception) {
+      throw CrdWebException("Unable to retrieve offender key dates", HttpStatus.NOT_FOUND)
```
james-baker-digital-justice, line 54: worth logging something do you think as original cause is lost? Also if the only thing that's throwing is the fold why not just throw CrdWebException straight away instead of catch and re-throw?

FaizJD030: The catch also covers error in other parts of the method, so i have logged the error before throwing the custom exception.

james-baker-digital-justice (approved):
LGTM

PR #675, opened 2024-05-10
Fix multiple offences to sentence analysed output

james-baker-digital-justice:
When you have multiple offences to a sentence and it is not the same as previous calculation then the normalised sentences were being lost.


joelstobart-moj (approved):
LGTM

PR #678, opened 2024-05-15
Add logic to calculate ESL for Manual Release date entries

FaizJD030:
https://dsdmoj.atlassian.net/browse/CRS-1937.

This PR addresses the calculation of ESL for manual entry journey and saves that to NOMIS after converting to Years Months and Days. When there is Indeterminate sentences or absence of SED then the release date is set to 0 Years, 0 Months and 0 days


joelstobart-moj (approved):
LGTM

PR #680, opened 2024-05-16
Crs 1943 Botus tests

dave-virden:
Creates the unit tests from the tickets plus a couple of others to check that taggged bail and remand are not applied to Botus sentences (this was noticed whilst testing the rest of the Botus work)


james-baker-digital-justice (approved):
LGTM

PR #682, opened 2024-05-17
CRS-1971

joelstobart-moj:
- Fixes to build string to only search the initialTypes rather than the whole ReleaseDateTypes.
- Correct spelling of initial


james-baker-digital-justice (approved):
LGTM

PR #684, opened 2024-05-17
Formatting of getReleaseDateType to match the rest of the file

joelstobart-moj:
Formatting of transformfunctions to use when syntax Simplifying the if as both sides appear to be identical


dave-virden (approved):
LGTM

PR #685, opened 2024-05-20
Fix bulk comparison with SDS exclusion lookup

james-baker-digital-justice:
We were combining all sentences and offences and doing a single lookup which could produce too many offences and make the MO URL too long.


joelstobart-moj (approved):
LGTM

PR #686, opened 2024-05-20
Extract calculation options from booking and feature toggles

james-baker-digital-justice:
Moved the calculate ersed flag and SDS early release flag into their own options object to be passed around for calcluation. This allows modification of the options for recalculation for SDS4x as well as improves cohesion of booking object.


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/Booking.kt
```diff
@@ -12,7 +12,6 @@ data class Booking(
   val returnToCustodyDate: LocalDate? = null,
   val fixedTermRecallDetails: FixedTermRecallDetails? = null,
   val bookingId: Long = -1L,
-  val calculateErsed: Boolean = false,
```
joelstobart-moj, line 15: after I did all that spelling correction ;)

james-baker-digital-justice: lol sorry

PR #689, opened 2024-05-21
CRS-1951: Default date logic for SDS early release

james-baker-digital-justice:
To allow pre-calculation of release dates before SDS early release takes effect we must adjust the release date to the commencement date if it would fall before the commencement date or the 50% date if they would have been release anyway before the new rules take effect.


srowlands-moj (approved):
Minor trigger at having Adjustment in the name of the `SDSEarlyReleaseAdjustmentService`

PR #697, opened 2024-05-28
Commit rollback change to the API client

joelstobart-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/ManageOffencesApiClient.kt
```diff
@@ -16,33 +18,42 @@ class ManageOffencesApiClient(@Qualifier("manageOffencesApiWebClient") private v
...
-          "PCSC indicator for offence lookup failed for $offencesList," +
-            " can not proceed to perform a sentence calculation",
-        )
+      .retryWhen(
+        Retry.backoff(5, Duration.ofMillis(100))
+          .maxBackoff(Duration.ofSeconds(3))
+          .doBeforeRetry { retrySignal ->
+            log.warn("getPCSCMarkersForOffences: Retrying [Attempt: ${retrySignal.totalRetries() + 1}] due to ${retrySignal.failure().message}. ")
+          },
+      )
```
srowlands-moj, line 33: Could potentially filter this down a bit rather than ALL errors but it's good for now

srowlands-moj (approved):
One comment, but it's whether we want to dig explicitly into the reasons for retry right now:
```               .retryWhen(Retry.backoff(5, Duration.ofSeconds(5))
                            .filter(throwable -> throwable instanceof HttpServerErrorException ||
                                    throwable instanceof WebClientRequestException && throwable.getCause() instanceof TimeoutException)
                    )``` as an example

PR #698, opened 2024-05-28
CRS-1982: Add retry and failure handling to bulk comparison

james-baker-digital-justice:
Retry several times to allow for intermittent network failures with external dependencies but after the retries are exhausted record the failure. Failures are returned for display in the UI alongside successful results


joelstobart-moj (approved):
LGTM

PR #705, opened 2024-05-30
Improve test coverage for Manage Offences retries

joelstobart-moj:
- Switch to an exception that extends RuntimeException
- Add integration tests


ldlharper (approved):
LGTM

PR #707, opened 2024-06-03
CRS-1988: Switch HDCED4+ to custodial period based

james-baker-digital-justice:
HDCED4+ is now calculated using half the custodial period instead one quarter of the sentence length. This provides tolerance for other release point multipliers being used, e.g., SDS4x


joelstobart-moj (approved):
LGTM

PR #709, opened 2024-06-04
CRS-1852 Add historical TUSED to BOTUS sentence.

dave-virden:
If a historical TUSED is present in NOMIS it will be added to the BOTUS sentence dates if it has not expired (is after the ARD/SED of the BOTUS senetence).


joelstobart-moj (approved):
LGTM

joelstobart-moj (approved):
LGTM

PR #710, opened 2024-06-06
CRS-1981: HDCED4+ SDS+ consec chain fix for remand and tagged bail

james-baker-digital-justice:
Apply remand and tagged bail to the SDS+ notional CRD before calculating HDCED4+. This is to ensure that it is not before the CRD of the SDS+ sentences.


joelstobart-moj (approved):
LGTM

PR #711, opened 2024-06-07
CRS-1995 Move the BulkComparison to use the system clients credentials.

dave-virden:
CRS-1995 Move the BulkComparison to use the system clients credentials.

HAAR have added RELEASE_DATE_MANUAL_COMPARER role to all the system clients see https://dsdmoj.atlassian.net/browse/HAAR-2716


joelstobart-moj (approved):
LGTM

PR #712, opened 2024-06-07
CRS-1997 Stop preventing HDCED when there are  EDS/SOPC sentences

srowlands-moj:
* Enables HDCED to be generated for chains that have EDS/SOPC in them, when it was previously omitted from the results.


joelstobart-moj (approved):
LGTM

PR #717, opened 2024-06-11
CRS-1817 Update alt-preprod to use normal calc params

srowlands-moj:
* Update alt preprod to use normal calc params
* Set hdc4 commencement date in the past


dave-virden (approved):
LGTM

PR #726, opened 2024-06-27
CRS-1993 Update a Calc reason from HDC to HDCAD

FaizJD030:
https://dsdmoj.atlassian.net/browse/CRS-1993


joelstobart-moj (approved):
LGTM

PR #727, opened 2024-06-27
CRS-1855 Store retrieved TUSED in bulk calc

dave-virden:
Uses the same logic as CRS-1852 to store get and store a TUSED from a previous booking and applies it if it is applicable.


FaizJD030 (approved):
LGTM

PR #729, opened 2024-07-03
CRS-2032 Prevent HDCED from being generated when booking has an SDS+ sentence

srowlands-moj:
* Change to align with how HDC4 policy is being applied by OMUs, any booking with a SDS+ Sentence is deemed ineligible for HDC


FaizJD030 (approved):
LGTM

PR #737, opened 2024-07-09
Test case from examples of tranch 1 logic in SDS40 document.

FaizJD030:


FaizJD030:
Addressed as part of another PR - https://github.com/ministryofjustice/calculate-release-dates-api/pull/751

PR #743, opened 2024-07-11
Use the identical "fourYearsOrMore" function and strip duplicate.

joelstobart-moj:
Business logic confirmed from ticket


srowlands-moj (approved):
LGTM

PR #748, opened 2024-07-12
Change alt envs around

srowlands-moj:
* Switch the alt and normal envs around so the normal envs have SDS Early Release and alt is for SDS not supported changes
* Sneaky code review comments from CRS-2035 changes that were trivial


joelstobart-moj (approved):
Good apart from the linting :)

joelstobart-moj (approved):
LGTM

PR #750, opened 2024-07-15
Turn off SDS unsupported feature in alt envs

srowlands-moj:
* Make alt-devs non SDS change to enable testing and progression of other non related tickets


joelstobart-moj (approved):
LGTM

PR #751, opened 2024-07-15
CRS-2041 Additional test cases to support the SDS40 tranching logic

FaizJD030:
https://dsdmoj.atlassian.net/browse/CRS-2041


src/test/resources/test_data/overall_calculation_response/custom-examples/crs-2041-SDS40-document-example1-tranche2.json
```diff
@@ -0,0 +1,9 @@
+{
+  "dates": {
+    "SLED": "2030-10-23",
+    "CRD": "2024-10-15",
```
srowlands-moj, line 4: If this is a defaulting case and we are using alt-3 params i'm a bit confused as to why this isn't 2024-10-22 🤔

srowlands-moj (approved):

PR #752, opened 2024-07-17
CRS-2054 Add terrorism exclusion and refactor to use updated endpoint

dave-virden:


dave-virden:
PLEASE DON'T MERGE, depends on Manage Offences API updates which will follow

FaizJD030 (approved):
LGTM

PR #755, opened 2024-07-19
CRS-2056 Return SDS40 tranche

srowlands-moj:
* Update detailed calc results to contain the allocated SDS40 tranche


joelstobart-moj (approved):
LGTM

PR #757, opened 2024-07-24
CRS-2062 Use standard TUSED when standard CRD is used

srowlands-moj:
* Ensure that when CRD is not defaulted and the standard CRD is used, then the standard TUSED is also adopted.


joelstobart-moj (approved):
LGTM

PR #758, opened 2024-07-25
CRS-2063 Add test to cover initial scenario

srowlands-moj:
* Additional RW test scenario for SDS40 with exclusion and FTR


joelstobart-moj (approved):
LGTM

PR #759, opened 2024-07-25
CRS-2065 Fix tranche not being carried through in all cases

srowlands-moj:
* Add RW test cases
* Correctly pass through tranche when there is no relevant defaulting applied


joelstobart-moj (approved):
LGTM

PR #774, opened 2024-07-31
CRS-2072 Recalls prior to SDS40 TUSED at 50%

srowlands-moj:
* Update defaulting logic to use the original calc TUSED if it's an SDS recall type with a CRD before SDS40 T1 commencement


joelstobart-moj (approved):
LGTM

PR #776, opened 2024-07-31
Update to SDS40

joelstobart-moj:
Add the spring profile for SDS40
Enable SDS Early Release
Disable unsupported


srowlands-moj:
Not required changes made under a previous PR and deployed.

PR #778, opened 2024-08-02
CRS-2076 and CRS-2074 multiple SDS40 calc changes

srowlands-moj:
* Update calculation service to explictly calc both variations so other decisions are made easier
* Through correct calc selection the validation issue in CRS-2076 should be resolved


joelstobart-moj (approved):
LGTM

PR #788, opened 2024-08-08
CRS-2081 Set Tranche to 0 when CRD is unchanged

srowlands-moj:
* Set tranche to 0 when the release date is unaffected
* Ensure that recall sentence length is only considered in the tranche decision if the SLED is on or after T2 commencement date
* Correctly remove term of imprisonment time from tranche calculation


dave-virden (approved):
LGTM

PR #789, opened 2024-08-13
CRS-2008 BOTUS consecutive to SDS and SDS consecutive to BOTUS

dave-virden:
Tests and alterations for various scenarios.


joelstobart-moj (approved):
LGTM

PR #794, opened 2024-08-19
CRS-2081 Fix consec bug

srowlands-moj:
* Fix for CRS-2081, minBy is actually minBy or null, so when all the sentences were discounted in the consec sentence, was throwing NPE


FaizJD030 (approved):
LGTM

PR #804, opened 2024-08-23
CRS-2081 Fix tranche allocation rules 

srowlands-moj:
* Don't adjust tranche to 0 if CRD is not defaulted
* Set the tranche to 0 if the adjusted dates are the same as the standard result


joelstobart-moj (approved):
LGTM

PR #811, opened 2024-08-30
Handle HDCED dates that wouldn't be allowed after the 40 but were at the 50% point

joelstobart-moj:
* Change to add the equivalent values (ReleaseDate and ReleaseDateDouble) to make the SentenceCalculation interface for HDCED and HDCED4 equivalent.

* Added the relevant test case
* refactored code to use the new SetenceCalculation interface
* introduced new logic to do a second check for a HDCED if the release point may be HDCED at the 50 and none is provided by the 40
* Passed the custodialPeriod into the calculateHdced functions rather than gathering them from the sentence calculation.
* Fixed linting issues


srowlands-moj (approved):
Only comment is we are probably at risk of bleeding SDS40 further out, I'm not saying it isn't necessary but I've been thinking about how to do this cleanly so it can be removed with ease at some point

joelstobart-moj:
Agree - will try and think of another way

PR #812, opened 2024-08-30
Identify HDCED dates from before commencement that only come from 50% release

joelstobart-moj:
Small tweak to how the defaulting service identifies whether a date… is before the tranche (if it comes only from the 50%)


srowlands-moj (approved):
LGTM

PR #815, opened 2024-09-03
Test framework changes before looking at the issue

joelstobart-moj:
* Introduce testing for the validation format
* Add the test files, but don't wire them in  - they'll fail in the pipeline, and I'd like to merge this test framework change
* Small linting and cleanup


srowlands-moj (approved):
LGTM

srowlands-moj (changes requested):
Looking at the failing check need to do a string comparison on the code type

srowlands-moj (approved):
LGTM

PR #820, opened 2024-09-06
CRS-2126 Ensure that only sentences sentenced before T1 are considered for tranching

srowlands-moj:
* Filter out any sentences sentenced post T1 from the tranching logic


joelstobart-moj (approved):
LGTM

PR #826, opened 2024-09-10
CRS-2077 Allow HDCED to be passed when SDS40 sentences are present

srowlands-moj:
* Pass through HDCED when SDS40 sentences are present such that it can be defaulted
* Update Defaulting service to remove HDCED from final dates if required (when PRRD is later)


joelstobart-moj (approved):
LGTM

PR #829, opened 2024-09-10
RCLL-108 Add endpoint to allow record-a-recall service to performs a transitory calculation

yasinmustafa:
The calculation will be performed and stored as a transitory calc (status =  RECORD_A_RECALL), therefore will not appear in the CRD UI or be published to NOMIS.


src/main/resources/migration/common/V60__record_a_recall_calculation_reason.sql
```diff
@@ -0,0 +1,2 @@
+-- These calcs are not pushed to nomis, but nomis_comment populated due to mandatory constraint
```
yasinmustafa, line 1: when performing a calc a  reason must be supplied, hence this new reason was inserted

ldlharper (approved):
LGTM

PR #837, opened 2024-09-17
override HDCED to the PRRD if the PRRD is later than the HDCED

joelstobart-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/SDSEarlyReleaseDefaultingRulesService.kt
```diff
@@ -84,27 +84,44 @@ class SDSEarlyReleaseDefaultingRulesService(
...
+          // See BookingExtractionService#extractMultiple
+          removeHDCEDDates(dates)
+        }
       }
     }
   }
 
+  private fun setHDCEDDates(date: LocalDate, dates: MutableMap<ReleaseDateType, LocalDate>) {
+    dates[ReleaseDateType.HDCED4PLUS] = date
+    dates[ReleaseDateType.HDCED] = date
```
srowlands-moj, line 118: Only thing I would say is this might cause a couple of problems until HDCED4PLUS has fully taken over from HDCED, as it is possible that there isn't a HDCED date... BUT in the real world it doesn't actually matter as only HDCED4PLUS is returned

srowlands-moj (approved):
LGTM

PR #838, opened 2024-09-17
introduce new breach of the 97 act error when the offence occurs afte…

joelstobart-moj:
…r 1997


joelstobart-moj:
Do no merge - messaging to be agreed

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/ValidationService.kt
```diff
@@ -381,6 +390,13 @@ class ValidationService(
     }
     return emptyList()
   }
+  private fun validateUnsupported97BreachOffencesAfter1Dec2020(sentencesAndOffence: List<SentenceAndOffenceWithReleaseArrangements>): List<ValidationMessage> {
+    val unSupportedEncouragingOffenceCodes = findUnsupported97BreachOffencesAfter1Dec2020(sentencesAndOffence)
+    if (unSupportedEncouragingOffenceCodes.isNotEmpty()) {
+      return listOf(ValidationMessage(UNSUPPORTED_BREACH_97))
+    }
+    return emptyList()
+  }
```
srowlands-moj, line 399: Could be return if but minor

PR #839, opened 2024-09-17
ADJST-833 Fixing UD for SDS40.

ldlharper:


src/test/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/resource/CalculationIntTest.kt
```diff
@@ -302,27 +302,6 @@ class CalculationIntTest(private val mockManageOffencesClient: MockManageOffence
     assert(!result.dates.containsKey(HDCED))
   }
 
-  @Test
```
ldlharper, line 305: All the changes in the file relate to having added the SDS40 config to `application-test.yml`.

I had to move a lot of the test data back 10 years so it didn't trigger early release. This test however was doing something strange with SDS+ and I'm not sure how it could've ever worked as it doesn't stub MO at all. I've removed it because this is covered both other tests in the file and other calculation tests.

src/test/resources/application-test.yml
```diff
@@ -76,6 +76,29 @@ feature-toggles:
 hdced4:
   hdc4-commencement-date: 2025-01-01
 
+release-point-multipliers:
```
ldlharper, line 79: All the integration tests now use `alt-3` params.

joelstobart-moj (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/resource/CalculationIntTest.kt
srowlands-moj: Possibly MO was being stubbed in base

ldlharper: Yea I realised now, its because I'd moved the dates for the test data back before SDS+ started! The SDS Int tests are already covered in this test class so I think its okay to remove it.

PR #842, opened 2024-09-20
Crs 2106 config driven support for inactive sentences

joelstobart-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/PrisonService.kt
ldlharper, line 1: I think logically this is all fine. My only comment would be that it might be better to have the function define the `supportInactiveSentencesAndAdjustments` as an override

Something like  `overrideSupportInactive: Boolean?`

Then only need one function for `getPrisonApiSourceData`

```
    return getPrisonApiSourceData(prisonerDetails,  overrideSupportInactive ?: featureToggles.supportInactiveSentencesAndAdjustments)

```

Then it makes it clear we're always using the environment config feature toggle unless its an endpoint that overrides it (relevant remand or recalls),

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/PrisonService.kt
```diff
@@ -24,16 +25,24 @@ import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.model.external.pris
...
-    val activeOnly = activeDataOnly || prisonerDetails.agencyId != "OUT"
-    return getPrisonApiSourceData(prisonerDetails, activeOnly)
+    return getPrisonApiSourceData(prisonerDetails, supportInactiveSentencesAndAdjustments)
   }
 
-  fun getPrisonApiSourceData(prisonerDetails: PrisonerDetails, activeDataOnly: Boolean = true): PrisonApiSourceData {
+  //  The activeDataOnly flag is only used by a test endpoint (1000 calcs test, which is used to test historic data)
+  fun getPrisonApiSourceData(prisonerId: String): PrisonApiSourceData {
+    val prisonerDetails = getOffenderDetail(prisonerId)
+    val supportInactiveSentencesAndAdjustments = featureToggles?.supportInactiveSentencesAndAdjustments ?: false
```
ldlharper, line 40: Why `featureToggles?` when the val is defined as non null?

joelstobart-moj: Because when we use the integration tests it's leaving the featuretoggles as null - and I didn't want to reinvent the tests...

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/PrisonService.kt
joelstobart-moj: I agree - but when the feature toggle is passed as a default, it needs to be wired into the parent class. However, if the parent class is mocked, there isn't an easy way of injecting the class for the default. It felt like a catch-22. 

Any alternative solutions welcome!

PR #845, opened 2024-09-23
CRS-2019 Reducing the duplicated consecutive sentences.

ldlharper:
There is already really good coverage for testing the functionality where a consecutive chain has multiple offences, some which are SDS plus and some which aren't. I could add a test case for the prod example. But I'm not sure it really adds anything to our coverage.


joelstobart-moj (approved):
LGTM

joelstobart-moj (approved):
LGTM

PR #855, opened 2024-09-25
Consolidate configurations

joelstobart-moj:
- Change application files to be more meaningful
- Remove excess ones
- Update files to reflect the new default (SDS40 and what is live) rather than the historic SDS at 50
- Update test cases where there are specific issues
- Add documentation to explain how developers should use the application profiles going forward.
- Make all operational commencement dates consistent in all files (some reflect speculative future dates, that have now passed).


joelstobart-moj:
NOTE: There is still an outstanding test case failure to be addressed. Pushed up for a partial review.

yasinmustafa (approved):
LGTM - nice work at tidying things up !

yasinmustafa (approved):
lgtm

PR #859, opened 2024-10-01
CRS-2135 Fix some some appilication of rules that feed into the hint texts on the UI

yasinmustafa:


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/SDSEarlyReleaseDefaultingRulesService.kt
```diff
@@ -214,7 +221,7 @@ class SDSEarlyReleaseDefaultingRulesService(
         releaseDate = commencementDate,
         unadjustedDate = early,
       )
-    } else if (standard != null && ((commencementDate != null && standard.isBefore(commencementDate)) || commencementDate == null)) {
+    } else if (standard != null && (commencementDate == null || standard.isBefore(commencementDate) || standard == early)) {
```
yasinmustafa, line 224: added an extra condition here + tweaked the logic to make it more concise

new condition `if standard==early` then standard takes precedence and those rules are enforced

The specific rule that relates it to standard  (as discussed with the analyst) is 'The CRD comes from a SDS with an excluded offence so the release date is calculated at 50% not 40%. '. Hiwever at this level we dont have access to that.. therefore gone with `standard == early`

src/test/resources/test_data/calculation_breakdown_response/custom-examples/crs-2047-sds-40-labels-5.json
```diff
@@ -55,7 +55,7 @@
       "unadjustedDate": "2027-04-22"
     },
     "CRD": {
-      "rules": ["SDS_EARLY_RELEASE_APPLIES"],
+      "rules": ["SDS_STANDARD_RELEASE_APPLIES"],
```
yasinmustafa, line 58: checked with the analyst (Cherky) on this and this test was wrong; i.e. it should actually be expecting SDS_STANDARD_RELEASE_APPLIES

joelstobart-moj (approved):
LGTM

PR #862, opened 2024-10-02
Fix bulk comparison

srowlands-moj:
* Amend transactional annotations such that calc request can be saved


joelstobart-moj (approved):
LGTM

PR #870, opened 2024-10-09
CRS-2156 Remove rogue filter on sentences causing incorrect defaulting

srowlands-moj:
* Correct sentence filtering logic that was causing some bookings to be incorrectly set to just standard release result
* Add RWE test cases


joelstobart-moj (approved):
Code change looks fine 
Not reviewing the examples (too much) 
LGTM

yasinmustafa (approved):
lgtm

PR #880, opened 2024-10-14
CRS-2157 fix failing test - fridaty release date is not applicable if release date is in the past

yasinmustafa:
This test was failing on Friday (11 Oct 2024) because the release date of the prisoner was 12 Oct 2024 (in the future on a Saturday) - so the 'non working day' logic was being applied and a hint saying release him on Friday was shown

But now the release date is in the past, so the hint will never be shown going forward (as per the logic in the code)


mg-moj (approved):
lgtm

PR #885, opened 2024-10-16
CRS-2157 Refactor SDSEarlyReleaseDefaultingRulesService to make it more readable

yasinmustafa:


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/SDSEarlyReleaseDefaultingRulesService.kt
```diff
@@ -29,144 +29,198 @@ class SDSEarlyReleaseDefaultingRulesService(
...
+    affectedBySds40 = (dates != standardReleaseResult.dates),
+  )
+
+  private fun getArdOrCrd(
     dates: MutableMap<ReleaseDateType, LocalDate>,
-  ): Pair<ReleaseDateType, LocalDate?> {
-    return if (dates[ReleaseDateType.ARD] != null) ReleaseDateType.ARD to dates[ReleaseDateType.ARD] else ReleaseDateType.CRD to dates[ReleaseDateType.CRD]
-  }
+  ): Pair<ReleaseDateType, LocalDate?> =
+    when {
```
mg-moj, line 106: This is great. You have to catch all possible outcomes with this approach and its easier to read. Especially with Sealed Classes.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/SDSEarlyReleaseDefaultingRulesService.kt
```diff
@@ -29,144 +29,198 @@ class SDSEarlyReleaseDefaultingRulesService(
...
   }
 
-  private fun removeHDCEDDates(dates: MutableMap<ReleaseDateType, LocalDate>) {
-    dates.remove(ReleaseDateType.HDCED)
+  private fun adjustPrrdForArdOrCrd(
+    dates: MutableMap<ReleaseDateType, LocalDate>,
+    ardOrCrdType: ReleaseDateType,
+    ardOrCrdDate: LocalDate?,
+  ) {
+    if (dates.containsKey(ReleaseDateType.PRRD)) {
```
mg-moj, line 159: Should this condition include ReleaseDateType.HDCED? I assume this is only applicable if PRRD is present.

yasinmustafa: good question, going by the assumption that the old code was correct then yes.. I havent changed any functionality just refactored. but agree it does sound a bit odd at this level, but there are some weird rules in CRD - so very pkausible. Ive pasted the old code below that did this bit.

```

      if (dates.containsKey(ReleaseDateType.PRRD)) {
        if (controllingDate.second?.isAfter(dates[ReleaseDateType.PRRD]) == true) {
          dates.remove(ReleaseDateType.PRRD)
        } else {
          dates.remove(controllingDate.first)

          // PRRD is later than any other release date, therefore can not have a HDCED
          // See BookingExtractionService#extractMultiple
          removeHDCEDDates(dates)
        }
      }
```

joelstobart-moj (approved):
LGTM

mg-moj (approved):
lgtm

PR #888, opened 2024-10-18
Create a local version of data-extraction-cron job

srowlands-moj:
* Enables us to define the resources for the job, to avoid an OOM in extraction

**A change suggestion is being made to https://github.com/ministryofjustice/hmpps-helm-charts/blob/main/charts/generic-data-analytics-extractor/values.yaml such that this local version will no longer be required**


srowlands-moj:
This was getting silly to get it functional, superseded by https://github.com/ministryofjustice/hmpps-helm-charts/pull/162/files

PR #889, opened 2024-10-18
RCLL-141 Throw specific exception

danbenton-mojdt:
Throw specific exception when no dates on an offence cause calculation to fail


mg-moj (approved):
lgtm

PR #890, opened 2024-10-18
Update version of data-extractor 

srowlands-moj:
* Update version of data-extractor 
* Specify resources for data-extractor 
* Add helm charts dir to git ignore (when running locally it pulls the dependencies down)


mg-moj (approved):
lgtm

PR #895, opened 2024-10-22
CRS-2162 The validation for the overlapping-remand-logic should use the adjustedDeterminateReleaseDate (rather than unadjusted)

yasinmustafa:
+ change how the range of sentences is determined (relates to fixing CRS-2082)
TODO Add explicit TDD's for the scenario in the ticket
(not done yet as the final release dates after the calculation finishes dont match the ticket, that will be investigated next)


joelstobart-moj (approved):
LGTM

PR #898, opened 2024-10-23
Update validation criteria for recall sentences based on tranche commencement date

joelstobart-moj:
Update validation criteria for recall sentences based on tranche commencement date

- Add a condition to fire a validation message only for recall sentences sentenced before the tranche commencement date. Sentences after this date will no longer cause the error to appear.
- Move relevant tests for this exception to a separate file for better organization.
- Add two new tests to cover scenarios where sentences are before and after the tranche commencement date.


srowlands-moj (approved):
LGTM

PR #902, opened 2024-10-24
CRS-2174 Fix issue with manual calc and sentence construction

srowlands-moj:
* Allow manual calc to not use the uninitialised sentence calc when constructing sentences for ESL purposes
* Catch any errors resulting in ESL generation and set ESL to 0


joelstobart-moj (approved):
LGTM

PR #906, opened 2024-10-25
Remove TrancheOne and TrancheTwo, consolidate into a single Tranche class

joelstobart-moj:
- Removed TrancheOne and TrancheTwo classes.
- Consolidated logic into a single Tranche class, parameterised by tranche type.


yasinmustafa (approved):
LGTM

PR #907, opened 2024-10-28
CRD-2168-sds-hide-hints-logic

mg-moj:
Suppress default date hint text if a SDS sentence has been imposed on or after 10/9/24

Add sentenceCalculationType and isSDSPlus to StandardDeterminateSentence and ConsecutiveSentencePart


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/ImportantDates.kt
```diff
@@ -14,4 +14,5 @@ object ImportantDates {
   val A_FINE_TEN_MILLION_FULL_RELEASE_DATE: LocalDate = LocalDate.of(2015, 6, 1)
   val SDS_DYO_TORERA_START_DATE: LocalDate = LocalDate.of(2005, 4, 4)
   val SOPC_TORERA_END_DATE: LocalDate = LocalDate.of(2022, 6, 28)
+  val SDS_CUT_OFF_DATE = LocalDate.of(2024, 9, 23)
```
srowlands-moj, line 17: Perhaps could be more descript in the name

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/CalculationResultEnrichmentService.kt
```diff
@@ -88,6 +92,24 @@ class CalculationResultEnrichmentService(
...
 
+  private fun showSDS40Hints(calculationBreakdown: CalculationBreakdown): Boolean {
+    if (!featureToggles.sdsEarlyReleaseHints) return false
+
+    if (calculationBreakdown.concurrentSentences.any {
+        !it.isSDSPlus && (it.sentenceCalculationType.isSdsCalcType() && it.sentencedAt.isAfter(SDS_CUT_OFF_DATE))
+      }
+    ) {
+      return false
+    }
```
srowlands-moj, line 103: Shame we don't have the sentence identification track here, would be a bit of a fudge getting it so agree what you've done is consistent 👍🏻

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/StandardDeterminateSentence.kt
```diff
@@ -12,6 +12,7 @@ data class StandardDeterminateSentence(
   override val offence: Offence,
   val duration: Duration,
   override val sentencedAt: LocalDate,
+  val sentenceCalculationType: String,
```
srowlands-moj, line 15: Might end up having to change a few tests :)

mg-moj: Yes, this will involve changing a lot of JSON test files. Although I could make it default to Null, even though I think its always present.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/CalculationResultEnrichmentService.kt
mg-moj: I missed this scenario. The description mentions excluding SDS40+, which I may have taken too literally. I'll examine the ticket again.

ldlharper:
Do we need to add the sentence calculation type to the model? If so why are we only adding it for SDS.

Can't you just do `if (sentence is StandardDeterminateSentence && sentence.recallType == null)`

That will get you any SDS sentence which is not a recall

PR #911, opened 2024-10-30
CRS-2185 Add STS21 type to SentenceCalc types and make it unsupported

srowlands-moj:
Small change to add the sentence calc type and make it unsupported for now


dave-virden (approved):
LGTM

PR #913, opened 2024-10-31
CRS-2190-add-dto-recall-enum

mg-moj:
Comparisons can sometimes contain DTO_RECALL validation codes. This was previously removed as its no longer needed.

It can still be used when indirectly parsing historic data.


srowlands-moj (approved):
LGTM

PR #916, opened 2024-11-05
Crs 2187 tused displayed in error

mg-moj:
Return historic TUSED if sentence is BOTUS and release date is after TUSED.

Do not calculate TUSED against Botus sentence where is does not apply.


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BookingExtractionService.kt
```diff
@@ -109,18 +110,20 @@ class BookingExtractionService(
...
-    if (sentence.isBotus()) {
-      if (sentence is BotusSentence && sentence.latestTusedDate != null && sentenceCalculation.expiryDate.isBefore(
-          sentence.latestTusedDate,
-        )
-      ) {
-        dates[TUSED] = sentence.latestTusedDate!!
-        historicalTusedSource = sentence.latestTusedSource!!
-      }
-    } else {
+    // exclude ESED for IIRC Botus sentence
```
srowlands-moj, line 113: IIRC (If I recall correctly) doesn't need to be here

srowlands-moj (approved):
LGTM

PR #917, opened 2024-11-05
CMG-182 Add things-to-do endpoint

yasinmustafa:
this is as part of some work in CCARD, to display a list of TODO's in every service

This is the initial placeholder endpoint and is subject to change depending on what the TODO item needs to display

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/ThingsToDoService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/ThingsToDoService.kt
@@ -1,10 +1,12 @@
 package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.service
 
 import org.springframework.stereotype.Service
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/ThingsToDoService.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    return ThingsToDo(
+      prisonerId = prisonerId,
+      thingsToDo = thingsToDo,
+    )
+  }
+
+  private fun isCalculationRequired(offenderDetails: PrisonerDetails): Boolean {
+    val sentencesAndOffences = sentenceAndOffenceService.getSentencesAndOffences(offenderDetails.bookingId)
+    // TODO fix bug that affects getAnalyzedAdjustments method - see CRS-2199
```
yasinmustafa, line 35: Will be looking at this bug as part of the wider piece of this TODO's work

joelstobart-moj (approved):
LGTM

PR #920, opened 2024-11-06
CRS-2175 Validate SHPO SX03 sentences and redirect to manual entry if required

srowlands-moj:
* Validate SDS40 considered SHPO offences meet the criteria for manual entry
* Validate that SHPO SX03 sentences post T1 are also routed to the manual entry journey


joelstobart-moj (approved):
LGTM

PR #921, opened 2024-11-07
Align SDS40 Consec Manual Routing feature switch across envs

srowlands-moj:
* Make feature switch for SDS40 consec manual routing consistent across envs


joelstobart-moj (approved):
LGTM

PR #924, opened 2024-11-11
CRS-2175 Ensure that the sentence CRD is after T1

srowlands-moj:
* Previously the check for CRD after T1 wasn't being conducted on a per sentence basis
* Add boundary case to expect the exception causing manual entry redirect to be thrown when CRD of SHPO sentence is before T1


mg-moj (approved):
lgtm

PR #925, opened 2024-11-11
CRS-2186-shpo-offence-violation

mg-moj:
validation to prevent release dates being calculated for prisoners where they have at least one SDS40 eligible SDS sentence imposed for a Sentencing Act 2020 (SA20) Breach of Sexual Harm Prevention (SHPO) order offence and the offence date is before 1/12/20.


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/SHPOValidationService.kt
```diff
@@ -0,0 +1,72 @@
...
+    (isSdsSentenceInValid(sentence) || isSdsHistoricSentenceInValid(sentence))
+
+  private fun isRelevantOffence(sentence: StandardDeterminateSentence): Boolean =
+    !sentence.isSDSPlus && shpoOffenceCodes.contains(sentence.offence.offenceCode)
+
+  private fun isConsecutiveSdsHistoricSentenceInValid(consecutiveSentence: ConsecutiveSentence): Boolean =
+    consecutiveSentence.sentencedAt.isBefore(SDS_40_COMMENCEMENT_DATE) &&
+      consecutiveSentence.sentenceCalculation.releaseDate.isAfter(SDS_40_COMMENCEMENT_DATE)
+
+  private fun isSdsSentenceInValid(sentence: CalculableSentence): Boolean =
```
srowlands-moj, line 64: My apologies if I caused this, but it was how the ticket was written....

So this function and the next can be combined into one and the detection of SDS40 commenecement removed - in both scenarios we are doing the same check, which is, is the SHPO offence dated before the SHPO commencement date.

srowlands-moj (changes requested):
I think you can ditch the SDS40 bit, again sorry if I added to the confusion there, but the ticket was worded like that for sure.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/SHPOValidationService.kt
```diff
@@ -0,0 +1,72 @@
...
+    sentence.sentencedAt >= SDS_40_COMMENCEMENT_DATE &&
+      sentence.offence.committedAt < SHPO_OFFENCE_COMMENCEMENT_DATE
+
+  private fun isSdsHistoricSentenceInValid(sentence: StandardDeterminateSentence): Boolean =
```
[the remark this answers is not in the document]
mg-moj: Good point, not sure why I named it this way now you point it out. I'll change it.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/ValidationService.kt
```diff
@@ -79,6 +80,7 @@ class ValidationService(
     messages += recallValidationService.validateUnsupportedRecallTypes(booking)
     messages += postCalculationValidationService.validateSDSImposedConsecBetweenTrancheDatesForTrancheTwoPrisoner(booking, calculationResult)
     messages += postCalculationValidationService.validateSHPOContainingSX03Offences(booking, calculationResult)
+    messages += shpoValidationService.validate(booking)
```
srowlands-moj, line 83: As you won't care about the releaseDate anymore this can move to happen before calc

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/SHPOValidationService.kt
srowlands-moj: After discussion with Scott this was intentional, in order to limit the scope of those affected by the change.
We do have other validation for SA 20 offences being incorrectly dated, but didn't want to have to do the impact analysis of number of sentences that it would impact now.

srowlands-moj (approved):

PR #936, opened 2024-11-19
CRS-2210 Fixing bug with concurrent UAL license recalls.

ldlharper:
Bug where UAL occurs before the overall CRD. but after a concurrent CRD. UAL should only apply to the concurrent sentence if its a recall and shouldn't affect its original determinate release date.


srowlands-moj (approved):
LGTM

PR #953, opened 2024-12-06
Suppress HDCED if equal to or later than CRD

joelstobart-moj:
- Updated `HdcedExtractionService` to suppress HDCED when it is equal to or later than CRD/ARD, ensuring compliance with AC1.
- Added utility methods `latestAdjustedReleaseDateIsAfterHdced` and `releaseDateIsAfterHdced` for precise HDCED handling.
- Modified `BookingExtractionService` to prevent HDCED display or submission under the specified conditions.
- Added test cases for CRS-2211 and CRS-2211-non-ora to validate the suppression logic.
- Included relevant test data reflecting adjustments and expected outcomes, ensuring scenarios like ADAs are correctly processed.


yasinmustafa (approved):
LGTM

PR #965, opened 2024-12-18
CRS-2233-set-sds40flag-for-multiple-sentences

mg-moj:
Apply affectedBySds40 flag to multiple consecutive sentences.

Check for any calculated dates which are affected by SDS40 affected sentences.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/SentenceCalculation.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/SentenceCalculation.kt
@@ -41,6 +41,27 @@ data class SentenceCalculation(
       }
     }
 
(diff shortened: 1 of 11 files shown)


joelstobart-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/SentenceCalculation.kt
```diff
@@ -41,6 +41,27 @@ data class SentenceCalculation(
       }
     }
 
+  fun getDateByType(type: ReleaseDateType): LocalDate? {
+    if (!breakdownByReleaseDateType.containsKey(type)) {
+      return null
+    }
+
+    return when (type) {
```
ldlharper, line 49: is ERSED missing here?

PR #966, opened 2024-12-20
CRS-2237 Turn off consec-concurrent botus functionality in preprod/prod , on everywhere else

yasinmustafa:


joelstobart-moj (approved):
LGTM

PR #968, opened 2025-01-07
CRS-2240-data-analytics-extractor-memory-limit

mg-moj:
Increase memory limit to 4Gb in light of larger DB size.

A more optimised solution will be added in due course to prevent future memory issues.


joelstobart-moj (approved):
LGTM

PR #974, opened 2025-01-14
Add test to cover list D excluded scenarios

joelstobart-moj:
Fix issue where items on List D (over 7 years) use the SDS+ date, rather than the PCSC date 
Note: this, sadly, introduces a dependency between lists and date ranges we have tried to seperate as much as possible within code. There doesn't seem to be an obvious way around that :(


mg-moj (approved):
lgtm

PR #977, opened 2025-01-20
CRS-2254 Mock call to nonFridayReleaseService to avoid hint texts changing over time in the tests

yasinmustafa:
this changes some of the hints because the variable one 'The Discretionary..' will no longer show (the condition has been mocked to return false every time)


joelstobart-moj (approved):
LGTM

PR #982, opened 2025-01-21
CRS-2225-exclude-adas-from-terms

mg-moj:
Additional Added Days should only be applied to sentences, not terms.

Terms include AFine, DetentionAndTrainingOrder and Botus.

Botus class cannot implement the Term interface, otherwise the calc engine fails to retrieve the last release date.

New integration tests for Each term adjustments.


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/timeline/TimelineCalculator.kt
```diff
@@ -85,11 +89,13 @@ class TimelineCalculator(
   fun setAdjustments(sentences: List<CalculableSentence>, sentenceAdjustment: SentenceAdjustments) {
     sentences.forEach { sentence ->
       var adjustments = sentenceAdjustment
-      if (sentence is AFineSentence) {
+
+      if (sentence is Term || sentence is BotusSentence || sentence.isDto() || sentence.isOrExclusivelyBotus()) {
```
mg-moj, line 93: Currently unable to implement Botus sentence with the Term interface. Doing so causes code within the AdjustmentValidationService to fail, as theres no release date. [https://github.com/ministryofjustice/calculate-release-dates-api/blob/main/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/AdjustmentValidationService.kt#L78](url)

ldlharper: Okay. I think if there is no release date then we should skip that validation. Then this could probably become

`if (sentence.sentenceParts().all { it is Term })`

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/AdjustmentValidationService.kt
```diff
@@ -75,9 +67,16 @@ class AdjustmentValidationService() {
   internal fun validateAdditionAdjustmentsInsideLatestReleaseDate(calculationOutput: CalculationOutput, booking: Booking): List<ValidationMessage> {
     val adjustments = getSortedAdjustments(booking)
     val latestReleaseDate = calculationOutput.custodialPeriod.mapNotNull { period ->
-      period.sentences.filterNot { it is Term }
+      period.sentences
+        .filterNot { it is Term }
         .maxOfOrNull { it.sentenceCalculation.releaseDateDefaultedByCommencement }
     }.maxOrNull()
+
+    if (latestReleaseDate !is LocalDate) {
```
ldlharper, line 75: Isn't this just `latestReleaseDate == null`?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/ValidationService.kt
```diff
@@ -80,7 +80,12 @@ class ValidationService(
     calculationOutput.custodialPeriod.forEach { messages += sentenceValidationService.validateSentenceHasNotBeenExtinguished(it) }
     messages += adjustmentValidationService.validateRemandOverlappingRemand(booking)
     messages += adjustmentValidationService.validateRemandOverlappingSentences(calculationOutput, booking)
-    messages += adjustmentValidationService.validateAdditionAdjustmentsInsideLatestReleaseDate(calculationOutput, booking)
+    if (!calculationOutput.sentences.all { it.isOrExclusivelyBotus() }) {
```
ldlharper, line 83: Is this necessary now that the null check is in place?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/AdjustmentValidationService.kt
```diff
@@ -75,9 +67,16 @@ class AdjustmentValidationService() {
   internal fun validateAdditionAdjustmentsInsideLatestReleaseDate(calculationOutput: CalculationOutput, booking: Booking): List<ValidationMessage> {
     val adjustments = getSortedAdjustments(booking)
```
ldlharper, line 68: If you want to be really explicit with this check

```
 val nonTermSentences = calculationOutput.custodialPeriod.flatMap { it.sentences }.filterNot { it is Term }
    if (nonTermSentences.isEmpty()) {
      return emptyList()
    }
    val latestReleaseDate = nonTermSentences.maxOf { it.sentenceCalculation.releaseDateDefaultedByCommencement }
   ```

ldlharper (commented):
I agree with the changes. I think it could be a bit tidier in the validation.

ldlharper (approved):
LGTM

PR #984, opened 2025-01-22
CRS-2199 Only consider confirmed calculations when comparing for calc…

ldlharper:
…ulation required thing to do.


joelstobart-moj (approved):
LGTM

PR #989, opened 2025-01-27
CRS-2212 Add post-tranche related adjustments to affected release date types in the timeline service

yasinmustafa:
TODO - there will be a follow up ticket that deals with the hint texts that are impacted by this
TODO - add breakdown specific unit tests, will be done in next ticket as well


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/timeline/BookingTimelineService.kt
```diff
@@ -209,6 +224,153 @@ class BookingTimelineService(
...
+  ): CalculationResult {
+    val applicableRules = calculation.breakdownByReleaseDateType[releaseDateType]?.rules
+
+    if (applicableRules?.contains(rule) == true) {
+      log.info("${releaseDateType.name} before adjustment: ${calculation.dates[releaseDateType]}")
+      val crd = calculation.dates[ReleaseDateType.CRD]!!
+
+      val numberOfDaysToAdjust = calculateAdjustmentDaysAfterCommencement(adjustments, commencementDate, crd)
+
+      if (numberOfDaysToAdjust != 0L) {
```
ldlharper, line 293: Would it make sense to do this check within the function above, before your start looping over every rule and release date type? The majority of time there won't be any adjustments after tranche commencement.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/timeline/BookingTimelineService.kt
```diff
@@ -209,6 +224,153 @@ class BookingTimelineService(
...
+
+  private fun updateBreakdownWithTrancheAdjustedRule(
+    calculation: CalculationResult,
+    releaseDateType: ReleaseDateType,
+  ): CalculationResult {
+    val existingBreakdown = calculation.breakdownByReleaseDateType[releaseDateType]
+    return calculation.copy(
+      breakdownByReleaseDateType = calculation.breakdownByReleaseDateType.plus(
+        releaseDateType to existingBreakdown!!.copy(
+          rules = existingBreakdown.rules.plus(ADJUSTED_AFTER_TRANCHE_COMMENCEMENT),
```
ldlharper, line 368: Could this be added as a `rulesWithExtraAdjustments` so we can see the number of days that the release date was adjusted by?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/timeline/BookingTimelineService.kt
```diff
@@ -209,6 +224,153 @@ class BookingTimelineService(
       .groupBy { it.date }
   }
 
+  private fun applyTrancheAdjustmentLogic(
```
ldlharper, line 227: Could this entire functionality be lifted into a new file?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/timeline/BookingTimelineService.kt
```diff
@@ -209,6 +224,153 @@ class BookingTimelineService(
...
+    adjustments: Adjustments,
+    commencementDate: LocalDate,
+    releaseDateType: ReleaseDateType,
+    rule: CalculationRule,
+  ): CalculationResult {
+    val applicableRules = calculation.breakdownByReleaseDateType[releaseDateType]?.rules
+
+    if (applicableRules?.contains(rule) == true) {
+      log.info("${releaseDateType.name} before adjustment: ${calculation.dates[releaseDateType]}")
+      val crd = calculation.dates[ReleaseDateType.CRD]!!
```
ldlharper, line 289: Is there always a CRD? Could we get here with an ARD or DTO release types? Maybe one to check with Cherky.

ldlharper: Or at least only apply this logic if the CRD is non-null.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/timeline/BookingTimelineService.kt
yasinmustafa: yes it would be more efficient - good spot

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/timeline/BookingTimelineService.kt
yasinmustafa: added a defensive check against the CRD being null... will consult Cherky on wheher CRD or DTO release types should also be considered. Will keep you posted but if any changes are required will do that in the follow up ticket that Cherky is drafting

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/timeline/BookingTimelineService.kt
yasinmustafa: yes make sense - done

In the next ticket (where this rule will be used for the hints) will add some unit tests to test the braekdown explicitly too

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/timeline/BookingTimelineService.kt
yasinmustafa: yes I think that would make sense

ldlharper (approved):
LGTM

PR #990, opened 2025-01-28
CRS-2196-record-reason-for-manual-journey

mg-moj:
On saving a manual calculation, re-run validation to obtain the list of validation messages resulting in the manual calculation.

Store each validation message and code against the calculation request.


ldlharper (approved):
LGTM

PR #1000, opened 2025-02-07
Crs 2137 updated ftr rules

mg-moj:
CRS-2137-updated-ftr-rules

Where a fixed term recall is issued during the licence period of an under 12 month sentence that is running parallel to the licence period of a 12 months and over sentence:

Later SLED from over 12 month sentence

Where there are concurrent sentences and the oﬀender is subject to licences from both an under 12 month sentence and an over 12 month sentence, and the SLED of the over 12 month sentence ends later than the under 12 month SLED:

Then a 28 day FTR should be issued

Release would be on the 28th day or the SLED of the over 12 month sentence whichever is earlier

Later SLED from under 12 month sentence

If the later SLED comes from the under 12 month sentence as long as there are 14 days or more to the SLED of the over 12 month sentence:

The recall will always be a 28 day FTR

Release would be on the 28th day or the SLED of the over 12 month sentence whichever is earlier

Later SLED from under 12 month sentence

If however there are less than 14 days to the SLED of the over 12 months sentence a 14 day FTR should be issued:

The prisoner will serve to the earlier of either the 14th day or the SLED of the under 12 month sentence

Update calculation service test to accept multiple expected exceptions.


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/timeline/BookingTimelineService.kt
```diff
@@ -189,7 +191,7 @@ class BookingTimelineService(
           }
           currentSentenceGroup.clear()
         }
-        latestCalculation = timelineCalculator.getLatestCalculation(releasedSentenceGroups.map { it.sentences }, offender)
+        latestCalculation = timelineCalculator.getLatestCalculation(releasedSentenceGroups.map { it.sentences }, offender, returnToCustodyDate)
```
mg-moj, line 194: Its not great having to pass around the return to custody date, but there appears no other way to have the date available within the BookingExtractionService

src/test/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/CalculationTransactionalServiceTest.kt
```diff
@@ -211,12 +212,15 @@ class CalculationTransactionalServiceTest {
       TestUtil.objectMapper().writeValueAsString(calculatedReleaseDates),
     )
     if (expectedValidationMessage != null) {
-      assertThat(returnedValidationMessages).hasSize(1) // etc
-      assertThat(returnedValidationMessages[0].code.toString()).isEqualTo(expectedValidationMessage)
+      val expectedExceptions = expectedValidationMessage.split("|")
```
mg-moj, line 215: This is to enable multiple exceptions and assertions

yasinmustafa (approved):
lgtm

PR #1005, opened 2025-02-12
CRS-2260 Add null check for TUSED breakdown - could be null via the full-validation route when a BOTUS exists

yasinmustafa:
Possibly some tech debt here because there are no integration/TDD tests that test this functionality via the full-validation route

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BookingExtractionService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BookingExtractionService.kt
@@ -589,14 +589,16 @@ class BookingExtractionService(
       .filter { it.sentenceCalculation.topUpSupervisionDate != null }
       .maxByOrNull { it.sentenceCalculation.topUpSupervisionDate!! }
 
-    return if (latestTUSEDSentence != null &&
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BookingExtractionService.kt
```diff
@@ -589,14 +589,16 @@ class BookingExtractionService(
...
 
-    return if (latestTUSEDSentence != null &&
-      latestTUSEDSentence.sentenceCalculation.topUpSupervisionDate!!.isAfter(
-        latestLicenseExpiryDate,
-      )
-    ) {
-      latestTUSEDSentence.sentenceCalculation.topUpSupervisionDate!! to latestTUSEDSentence.sentenceCalculation.breakdownByReleaseDateType[TUSED]!!
-    } else {
-      null
+    return latestTUSEDSentence?.let { sentence ->
```
yasinmustafa, line 592: refactored this a little bit to ease readability, but really just added a null check

mg-moj: Agree it's nicer using the let scope

mg-moj (approved):
lgtm

PR #1008, opened 2025-02-17
crs-2137-enable-revised-ftr-rules

mg-moj:
Enable revised FTR rules for production environment.


yasinmustafa (approved):
lgtm -  I prefer to remove the feature toggle altogether if we're enabling in live - but not too precious about it.

PR #1013, opened 2025-02-21
CRS-2259 Only show 'TUSED from historical source' hint text if the tused date used is actually from the historical source

yasinmustafa:
**Before** 
If an active BOTUS sentence exists
AND a previous TUSED has been calculated
Then show hint 'TUSED from last calculation saved to NOMIS'

**New**
If an active BOTUS sentence exists
AND a previous TUSED has been calculated
AND the previous TUSED is being used for this calculation (i.e. the previous calculated date > than the new calculated TUSED)
Then show hint 'TUSED from last calculation saved to NOMIS'


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/SentenceAdjustedCalculationService.kt
```diff
@@ -248,6 +248,7 @@ class SentenceAdjustedCalculationService(
   private fun getBotusTusedDate(sentence: BotusSentence, sentenceCalculation: SentenceCalculation) {
     if (sentence.latestTusedDate?.isAfter(sentenceCalculation.releaseDate) == true) {
       sentenceCalculation.topUpSupervisionDate = sentence.latestTusedDate
+      sentenceCalculation.breakdownByReleaseDateType[TUSED] = tusedCalculator.getCalculationBreakdownForBotus(sentenceCalculation)
```
yasinmustafa, line 251: I think the method where this code sits would be better placed in the TusedCalculator class + there are other bits in here that do TUSED specific calc/breakdown stuff that I think should be moved. 

I didnt do it as part of this PR though, will discuss and perhaps do as a tech debt ticket

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/TusedCalculator.kt
```diff
@@ -80,6 +80,14 @@ class TusedCalculator(val workingDayService: WorkingDayService) {
     )
   }
 
+  fun getCalculationBreakdownForBotus(sentenceCalculation: SentenceCalculation): ReleaseDateCalculationBreakdown =
+    ReleaseDateCalculationBreakdown(
+      rules = setOf(CalculationRule.BOTUS_LATEST_TUSED_USED) + if (sentenceCalculation.isImmediateRelease()) setOf(CalculationRule.IMMEDIATE_RELEASE) else emptySet(),
+      adjustedDays = getAdjustedDays(sentenceCalculation),
+      releaseDate = sentenceCalculation.topUpSupervisionDate!!,
```
mg-moj, line 87: As the method is public, it might be worth passing in the topUpSupervisionDate as a mandatory argument. Doing so removes the possibility of passing in a sentenceCalculation with a null topUpSupervisionDate

yasinmustafa: Good point @mg-moj , I think I'd prefer tp do it as tech debt on CRS-2287 though which covers this whole TUSED calculator, reason being that although the method is new code, it's actually pretty much a copy of the method above, and that has the same problem

CRS-2259 is about shifting some of the calling functionality into the TusedCalculator, and I think this would be better fixed with that .

PR #1015, opened 2025-02-25
Crs 2289 prevent logging no offences found

mg-moj:
Create new Exception for No Active Booking. 
Create complementary Handler with no log entries.


ldlharper (approved):
LGTM

PR #1018, opened 2025-02-26
crs-2262-concurrent-botus-feature-toggle

mg-moj:
Add FeatureToggle to calculate sentences where there is 1 or more active BOTUS Concurrent with other supported sentences / sentence chains.

Check Nomis sentence chains where the first element in the chain may be Botus but has no consecutiveToSequence value.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/BotusValidationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/BotusValidationService.kt
@@ -1,13 +1,14 @@
 package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.validation
 
-import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.config.FeatureToggles
+import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.model.SentenceAndOffenceWithReleaseArrangements
 import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.model.external.PrisonApiSourceData
 import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.model.external.SentenceAndOffence
 import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.model.external.SentenceCalculationType
 import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.model.external.SentenceCalculationType.BOTUS
 import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.validation.ValidationCode.BOTUS_CONSECUTIVE_OR_CONCURRENT_TO_OTHER_SENTENCE
+import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.validation.ValidationCode.BOTUS_CONSECUTIVE_TO_OTHER_SENTENCE
 
 @Service
 class BotusValidationService(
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/BotusValidationService.kt
```diff
@@ -1,33 +1,47 @@
...
+      val botusSentences = getBotusSentences(sourceData)
+
+      if (botusSentences.isEmpty() || botusSentences.size == sourceData.sentenceAndOffences.size) {
+        return emptyList()
+      }
 
-    if (botusSentences.isEmpty() || botusSentences.size == sourceData.sentenceAndOffences.size) {
-      return emptyList()
+      return listOf(ValidationMessage(code = BOTUS_CONSECUTIVE_OR_CONCURRENT_TO_OTHER_SENTENCE))
     }
```
srowlands-moj, line 43: It might be because i'm late to the party here, but this would prevent BOTUS concurrent to any other sentence, as the change to the validation code implies, but that's not how I read "This will simply be looking to calculate sentences where there is 1 or more active BOTUS Concurrent with other supported sentences / sentence chains." Do we want to be doing this when the concurrent journey is toggled on

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/BotusValidationService.kt
```diff
@@ -1,33 +1,47 @@
...
 ) {
 
   internal fun validate(sourceData: PrisonApiSourceData): List<ValidationMessage> {
-    if (featureToggles.botusConsecutiveJourney) {
-      log.info("BOTUS consecutive and concurrent journeys are feature enabled - therefore the 'unsupported' validation for BOTUS will not run")
-      return emptyList()
+    if (featureToggles.botusConcurrentJourney) {
+      val consecutiveSentences = sourceData.sentenceAndOffences
+        .filter { it.consecutiveToSequence != null }
+
```
yasinmustafa, line 22: not sure if I'm right on this but I thought both feature toggles could be on at the same time; I may be wrong though, worth checking with Katie

With this new code it looks like if CONCURRENT is on then it will always fail the consec check even if the CONSEC feature toggle is on

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/BotusValidationService.kt
mg-moj: That's correct. This is the existing validation code sat behind a feature toggle. 
The current implementation requires one or the other to be enabled, not both. I'll check if both feature toggles could be enabled.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/BotusValidationService.kt
```diff
@@ -36,7 +50,29 @@ class BotusValidationService(
...
+    val sentence = sourceData.sentenceAndOffences.firstOrNull { it.sentenceSequence == index }
+    return sentence?.let { isBotusSentence(it) } == true
+  }
+
+  private fun getConsecutiveChains(
+    consecutiveSentences: List<SentenceAndOffenceWithReleaseArrangements>,
+  ): List<List<Int>> {
+    val sentenceChains = mutableListOf<MutableList<Int>>()
+
+    for (sentence in consecutiveSentences) {
```
ldlharper, line 67: Does this assume that the sentences are in order?

ldlharper (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/BotusValidationService.kt
mg-moj: It does. If the prison API can return sentences out of order, I can add a sort

ldlharper: Yea they're not always in order.

srowlands-moj (approved):
LGTM

ldlharper (approved):
LGTM

PR #1021, opened 2025-02-28
crs-2262-crs-2292-add-primary-key-approved-dates

mg-moj:
approved_dates_submission_approved_dates doesn't have a primary key - add auto incremented id column


joelstobart-moj (approved):
LGTM

PR #1025, opened 2025-03-05
CRS-2297 Start of refactoring bulk comparison.

ldlharper:
Start of refactoring. Just creating an interface for bulk comparison and a bit of renaming/splitting. Then will introduce a new service that will use SQS instead.


joelstobart-moj (approved):
LGTM

PR #1044, opened 2025-03-11
CRS-2297 More bulk fixes where preprod refresh removed an existing co…

ldlharper:
…mparison and a fatal error.

Fatal errors now only errors which occur within the engine. All the lookups will use SQS DLQ retries.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BulkComparisonEventService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BulkComparisonEventService.kt
@@ -136,7 +136,7 @@ class BulkComparisonEventService(
         calculationUserInput,
         bulkCalculationReason,
         sourceData,
-        CalculationStatus.TEST,
+        CalculationStatus.BULK,
         username,
       )
     } catch (e: Exception) {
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BulkComparisonEventService.kt
```diff
@@ -122,14 +128,26 @@ class BulkComparisonEventService(
...
-    )
+    val establishmentValue = getEstablishmentValueForComparisonPerson(comparison, establishment)
+
+    val validationResult = try {
+      calculationTransactionalService.validateAndCalculate(
+        personId,
+        calculationUserInput,
+        bulkCalculationReason,
+        sourceData,
+        CalculationStatus.TEST,
```
srowlands-moj, line 139: Unrelated, but did we settle on TEST?

srowlands-moj:
Out of interest what's the DLQ retry limit?

srowlands-moj (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BulkComparisonEventService.kt
ldlharper: TEST is what we used to have when bulk testing was all done on the frontend. But could add a BULK value

ldlharper:
> Out of interest what's the DLQ retry limit?

There is no limit, it will just keep retrying forever. It will alert us in slack though

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BulkComparisonEventService.kt
srowlands-moj: Nice

srowlands-moj (approved):
LGTM

srowlands-moj (approved):
LGTM

PR #1059, opened 2025-03-25
Improving bulk comparison

ldlharper:
1. Increase timeout for getting prisoners at establishment.
2. Bulk message request only allows 10 messages at a time
3. Add new state for SETUP.
4. Mark comparison as error if exception happens in the setup


joelstobart-moj (approved):
LGTM

PR #1064, opened 2025-04-01
Crs 2283 removed duplicate consec sentences

mg-moj:
Check for none Consecutive sentences with consecutiveSentenceUUIDs.

Such sentences have likely been removed from an earlier consec sentence to avoid duplicates consec chains.

If the UUID is present for any current consec sentence chain, the sentence has more than one child  .

### DIFF
--- a/src/test/resources/test_data/overall_calculation/consecutive-sentences/crs-2283-ac5-5.json
+++ b/src/test/resources/test_data/overall_calculation/consecutive-sentences/crs-2283-ac5-5.json
@@ -1,0 +1,310 @@
+{
+  "offender": {
+    "reference": "ABC123",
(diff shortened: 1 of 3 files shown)


src/test/resources/test_data/overall_calculation/consecutive-sentences/crs-2283-ac5-5.json
```diff
@@ -0,0 +1,310 @@
+{
```
ldlharper, line 1: I'd suggest using the `BookingAnonymiser` for all your test cases, it makes them much easier to read.

ldlharper (approved):
LGTM

PR #1073, opened 2025-04-10
Set the CPU request limit to 1 so CPU isn't overly throttled

srowlands-moj:
* Extract failed on Wed morning due to CPU limits, it did pass again the following day however the CPU was throttled due to the initial request being fairly low but within limits


mg-moj (approved):
lgtm

PR #1078, opened 2025-04-14
crs-2333-add-post-calc-manual-journey-reasons

mg-moj:
Store post calculation validation errors resulting in manual journey.

Errors can be from pre-calculation or post-calculation.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/ManualCalculationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/ManualCalculationService.kt
@@ -22,6 +22,7 @@ import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.model.external.Upda
 import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.repository.CalculationOutcomeRepository
 import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.repository.CalculationReasonRepository
 import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.repository.CalculationRequestRepository
+import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.validation.ValidationCode
 import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.validation.ValidationService
 import java.time.LocalDate
 import java.time.Period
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/ManualCalculationService.kt
```diff
@@ -94,14 +96,34 @@ class ManualCalculationService(
...
+      if (preCalcManualJourneyErrors.isNotEmpty()) {
+        savedCalculationRequest.manualCalculationReason = preCalcManualJourneyErrors.map { transform(savedCalculationRequest, it) }
         calculationRequestRepository.save(savedCalculationRequest)
+      } else {
+        val calculationOutput = calculationService.calculateReleaseDates(
+          booking,
+          calculationUserInputs,
+        )
+
+        val postCalcManualJourneyErrorTypes = listOf(
```
ldlharper, line 113: Could we check if the returned validation code is has a `validationType` of `MANUAL_ENTRY_JOURNEY_REQUIRED` or any of the `UNSUPPORTED` types. Then we don't need to maintain a list here if we add more unsupported post calc types.

src/test/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/ManualCalculationServiceTest.kt
```diff
@@ -584,6 +587,41 @@ class ManualCalculationServiceTest {
     assertThat(actualRequest.calculationType).isEqualTo(CalculationType.MANUAL_DETERMINATE)
   }
 
+  @Test
+  fun `Check manual journey records validation reasons for post calculation errors`() {
```
ldlharper, line 591: Are there any Int tests for manual entry?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/ManualCalculationService.kt
mg-moj: Good suggestion, I'll update the code

ldlharper (approved):
LGTM

PR #1082, opened 2025-04-22
crs-2345-hdced-date-check

mg-moj:
Guard against null homeDetentionCurfewEligibilityDate while checking for conflicting sentences within the HdcedExtractionService.kt


joelstobart-moj (approved):
LGTM

PR #1083, opened 2025-04-23
Crs 2339 one day sentences

mg-moj:
One day standard determinate sentences should only have SED and ARD in place of SLED and CRD.


srowlands-moj (approved):
LGTM

PR #1085, opened 2025-04-24
crs-2351-bulk-calc-validation

mg-moj:
Ignore CONCURRENT_CONSECUTIVE_SENTENCES for bulk calc. Validation does not prevent calculations and is used for informational purposes.


joelstobart-moj (approved):
LGTM

PR #1088, opened 2025-04-25
crs-2283-move-to-pre-calc

mg-moj:
Refactor validation highlighting multiple consecutive chains against the same sentence.

Move validation to pre calc data.

Remove sentences derived from splitting a sentence with multiple offences into different sentences. Resulting sentences share the same sentenceSequence, sentenceSequence and caseSequence


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/SentenceValidationService.kt
```diff
@@ -52,6 +60,83 @@ class SentenceValidationService(
...
+
+    val affectedSentenceChains = childSentences.associate { childSentence ->
+      val chain = mutableListOf<ValidateConsecutiveSentenceRecord>().apply {
+        add(childSentence)
+        walkConsecutiveSentence(consecutiveSentences, childSentences = this)
+      }
+      childSentence.sentenceSequence to (chain.flatMap { it.terms } + rootSentence.terms)
+    }
+
+    val longestTerm =
```
ldlharper, line 98: Is this right? Wouldn't 12 months be longer than 2 years here?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/SentenceValidationService.kt
```diff
@@ -52,6 +60,83 @@ class SentenceValidationService(
     }
   }
 
+  private fun validateConsecutiveSentenceUniqueWithDuration(sentences: List<SentenceAndOffence>): List<ValidationMessage> {
+    val consecutiveSentences = sentences
```
ldlharper, line 64: I'm not sure exactly what this part is doing?

Looking to see if there are any consecutive sentences? Why check the case sequence and terms?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/SentenceValidationService.kt
```diff
@@ -52,6 +60,83 @@ class SentenceValidationService(
...
+
+    val duplicateConsecutiveSequences = consecutiveSentences
+      .map { it.consecutiveToSequence }
+      .groupBy { it }
+      .filter { it.value.size > 1 }
+      .values
+      .flatten()
+      .takeIf { it.isNotEmpty() } ?: return emptyList()
+
+    val rootSentence = sentences.firstOrNull { it.sentenceSequence == duplicateConsecutiveSequences.first() } ?: return emptyList()
```
ldlharper, line 87: Does this always work?

If you have:
2 consec to 1
3 consec to 2
4 consec to 2

The duplicated consecutive sequence will be sequence 2. But thats not the root sentence.

ldlharper:
Also can you add some Integration tests for this please?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/SentenceValidationService.kt
mg-moj: Yes you're right. I forgot to update this code after changing the logic elsewhere = from an earlier version. I'll update accordingly.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/SentenceValidationService.kt
mg-moj: The distinct filter removes any sentences with the same details. Sentences resulting from multiple offences against one Nomis sentence share the same information. Not removing them can trigger the validation by accident, as each sentence can have the same consecutiveToSequence value. The terms are used later to compute the total duration.

ldlharper: Oh I see, I didn't realise they'd already been flattened. I think they did that for SDS+. You should only need to use the sentence sequence to get distinct sentences.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/util/Consecutil.kt
```diff
@@ -0,0 +1,49 @@
+package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.util
+
+class Consecutil {
```
srowlands-moj, line 3: Pernickety: surprised the linter didn't complain it's not a capital U

srowlands-moj (approved):
LGTM, minor comment

PR #1099, opened 2025-05-13
update-dependencies-130525

mg-moj:
Update to HMPPS spring boot 8.1.0

Pin hibernate to version 6.5.3 to prevent transaction issues when using JPA and data classes.


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/BankHolidayService.kt
```diff
@@ -44,7 +45,7 @@ class BankHolidayService(
     }
   }
```
mg-moj, line 47: Method can fail when called from the timeline service. A read only transaction exception is thrown. The annotation requires Hibernate to use a new transaction which defaults to writable.

PR #1102, opened 2025-05-19
flyway-logging-level

mg-moj:
Set flyway logging level to "warn"
Remove real names from test / informational data
Replace MockBean with MockitoBean within affected tests


joelstobart-moj (approved):
LGTM

PR #1106, opened 2025-05-29
CRS-2368 Update to latest version of data-analytics-extractor

srowlands-moj:
* PR to test the upgrade to the latest version of data-analytics-extractor
* If this does not successfully run then this PR will need to be reverted prior to any other deployment to prod


joelstobart-moj (approved):
LGTM

PR #1126, opened 2025-06-27
RCLL-163 Include previous booking in calculation.

ldlharper:


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/AdjustmentsSourceData.kt
```diff
@@ -0,0 +1,37 @@
+package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.model
+
+import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.adjustmentsapi.model.AdjustmentDto
+import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.model.external.prisonapi.BookingAndSentenceAdjustments
+
+data class AdjustmentsSourceData(val prisonApiData: BookingAndSentenceAdjustments? = null, val adjustmentsApiData: List<AdjustmentDto>? = null) {
```
ldlharper, line 6: Previously were using a EIther to capture the adjustments data being from either prison-api or adjustments-api. But its really used for capturing success and error eithers. So I've refactored that as part of this change.

PR #1138, opened 2025-07-08
log booking JSON when preliminary and current data hashes differ

joelstobart-moj:
This helps identify changes in booking data that cause validation to fail when confirming a preliminary calculation. Logs are set to INFO temporarily for visibility


james-baker-moj (approved):
LGTM

PR #1140, opened 2025-07-09
CRS-2396 Early Release refactor. Allow mutilple early release configs with tra…

ldlharper:
…nches.

Potential issues with this implementation.

1. I've left SDS40 hard coded in place for validations and hint text. This would need implementing for SDS33 too if required.
2. The way the configuration is loaded for the csv file tests is a bit painful. I'd like to refactor this to be a spring test.
3. There is a file added for SDS33, the values are all made up.


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/CalculableSentence.kt
```diff
@@ -177,4 +177,10 @@ interface CalculableSentence {
 
   @JsonIgnore
   fun sentenceParts(): List<AbstractSentence>
+
+  @JsonIgnore
+  fun isAffectedByEarlyRelease(): Boolean = this.sentenceParts().any {
+    it.identificationTrack == SentenceIdentificationTrack.SDS &&
+      this.sentenceCalculation.unadjustedReleaseDate.findMultiplierBySentence(it) != 0.5
```
james-baker-moj, line 184: This feels like it might be a bug one day... could there be a more explicit way of determining early release?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/SDSEarlyReleaseExclusionType.kt
```diff
@@ -19,6 +19,9 @@ enum class SDSEarlyReleaseExclusionType {
   TERRORISM_T3,
   MURDER_T3,
   NO,
+  ;
+
+  fun isSDS40Tranche3Exclusion(): Boolean = this.name.endsWith("_T3")
```
james-baker-moj, line 24: It feels safer to use a flag on the enum instead of relying on the name but I suppose it's pretty likely this will require further updates for 33.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/timeline/BookingTimelineService.kt
```diff
@@ -119,24 +122,24 @@ class BookingTimelineService(
           offender,
           returnToCustodyDate,
         ).copy(
-          sdsEarlyReleaseAllocatedTranche = trancheAndCommencement.first,
-          sdsEarlyReleaseTranche = trancheAndCommencement.first,
+          sdsEarlyReleaseAllocatedTranche = allocatedTranche?.let { SDSEarlyReleaseTranche.fromDate(it.date, earlyReleaseConfigurations) } ?: SDSEarlyReleaseTranche.TRANCHE_0,
+          sdsEarlyReleaseTranche = allocatedTranche?.let { SDSEarlyReleaseTranche.fromDate(it.date, earlyReleaseConfigurations) } ?: SDSEarlyReleaseTranche.TRANCHE_0,
```
james-baker-moj, line 126: Could we keep hold of the whole early release config rather than just the date? Or some type with name + date? Also could avoid the duplicate lookup here

james-baker-moj (dismissed):
A few minor comments but LGTM

mg-moj (dismissed):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/timeline/BookingTimelineService.kt
ldlharper: Agreed, I've refactored this service to use the tranche class rather than just passing a date around.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/SDSEarlyReleaseExclusionType.kt
ldlharper: This logic existed already, I've just moved it. I agree though, I'll make the change.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/CalculableSentence.kt
ldlharper: This is only used to set a db value used by analytics to see how many calcs are affected by SDS40. i've renamed it and made the logic SDS40 specific.

PR #1141, opened 2025-07-10
crs-2370-booking-data-error

mg-moj:
Ignore CONCURRENT_CONSECUTIVE validation messages when confirming calculation.

The validation type is informational rather than preventing a calculation.

Future refactoring should resolve such ambiguity.


joelstobart-moj (approved):
LGTM

PR #1149, opened 2025-07-16
CRS-2386: Update HDC rules regarding UAL (part 2)

james-baker-moj:
Further changes to the rules regarding UAL have been made, the 3rd calculation including UAL was not required as we only need to exclude UAL for any case where we default to the sentence date + 14 days.

Now the rules are as follows:

Step 1: Calculate HDC180 & HDC365 with Remand & TB If date falls before SD + 14, default to SD + 14 days for both HDC180 & HDC365

Step 2: Apply applicable UAL to calculated HDC180 and HDC365 dates

Step 3: Follow one of the following:
3a: If HDC180 falls before 03/06/2025 - Retain HDC180
3b: If HDC365 falls after 03/06/2025 Retain HDC365
3c: If HDC180 falls after 03/06/2025 and HDC365 is before 03/06/2025 Default to 03/06/25

Step 4: Apply ADAs & RADAs

Step 5: Defer any HDCED to later PRRD, PED, CRD, ARD of other sentences


mg-moj (approved):
lgtm

PR #1151, opened 2025-07-17
CRS-2401: Fix issue with unsupported sentences (attempt 2)

james-baker-moj:
Perform all pre-calculation validation in the /supported-validation API and immediately return without doing the manual entry journey checks if there are other validation errors.

This validation is performed again later and is displayed with custom messages so no need to return it here.


joelstobart-moj (approved):
LGTM

PR #1162, opened 2025-07-25
CRS-2399: Switch remaining builds to github

james-baker-moj:
We decided to remove the non-standard lint of open api docs as it wasn't providing any value and they are generated by Spring anyway.


ldlharper (approved):
LGTM

PR #1164, opened 2025-07-28
CRS-2381: Apply ERS30 to all tests

james-baker-moj:
Enabled the feature toggle in all tests and apply updates to the expected results. All checked with BAs to confirm the expected dates.


mg-moj (approved):
lgtm

PR #1184, opened 2025-09-02
CRS-2472: Consec spanning ERS30 commencement date fix

james-baker-moj:
When there is a consecutive sentence with sentences both before and after ERS30 commencement date, we do not need to calculate the ERS50 date and can just use the ERS30 calculation. No additional logic for single sentences after ERS30 is required as they will always generate a date after commencement by nature.


mg-moj (approved):
lgtm

PR #1187, opened 2025-09-04
Crs 2437 check for last last manual calc

mg-moj:
Check if last manual calculation is the same input data as the current Nomis booking.


ldlharper:
Can you write an Int test for this? Its good to make sure the JQL queries are tested.

PR #1211, opened 2025-09-24
Crs 2480 pss repeal

mg-moj:
Default ANY Calculated OR Retrieved TUSEDs that are ON OR AFTER the date of repeal to the day before.

Retain ANY Calculated or Retrieved TUSEDs that are BEFORE the date of repeal.

Not Calculate any NEW TUSED’s for sentences imposed ON or AFTER the date of repeal.


src/test/resources/test_data/overall_calculation/tused/crs-2480-ac1-1.json
ldlharper, line 1: Can the directory be named after tused repeal?

PR #1217, opened 2025-09-29
CRS-2414: New genuine override endpoint

james-baker-moj:
Also removed the old genuine override endpoints and the SST flags which were unused in the UI.


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/entity/CalculationType.kt
```diff
@@ -5,6 +5,5 @@ enum class CalculationType {
   MANUAL_DETERMINATE,
   MANUAL_INDETERMINATE,
   CALCULATED_WITH_APPROVED_DATES,
-  MANUAL_OVERRIDE,
-  CALCULATED_BY_SPECIALIST_SUPPORT,
```
james-baker-moj, line 9: these seem to be hangovers from previous attempts at GO and were unused by the UI

ldlharper (approved):
LGTM

PR #1218, opened 2025-09-30
CRS-2511: ERS defaulting to commencement date before sentence date

james-baker-moj:
We should only use ERS30 calculation if the sentence date is after the commencement date but also the logic that defaults to sentence date was still using the unadjusted ERSED to check vs the commencement date rather than the defaulted-to-sentence-date-ERSED .


ldlharper (approved):
LGTM

PR #1221, opened 2025-10-01
CRS-2414: Fix genuine override with no further detail

james-baker-moj:
Should have been a nullable type on the request object.


mg-moj (approved):
lgtm

PR #1233, opened 2025-10-07
RCLL-24 Use granualar role for recalls endpoint (used for calculations)

yasinmustafa:


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/resource/RecordARecallController.kt
```diff
@@ -21,7 +21,7 @@ class RecordARecallController(
   private val recordARecallService: RecordARecallService,
 ) {
   @PostMapping(value = ["/record-a-recall/{prisonerId}"])
-  @PreAuthorize("hasRole('RECORD_A_RECALL')")
+  @PreAuthorize("hasAnyRole('RECORD_A_RECALL', 'ROLE_CALCULATE_RELEASE_DATES__RECALL__CALCULATE__RW')")
```
yasinmustafa, line 24: The RECORD_A_RECALL role will become redundant once Ive refactored the recalls side.. will come back and remove it once done

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/resource/RecordARecallController.kt
```diff
@@ -21,7 +21,7 @@ class RecordARecallController(
   private val recordARecallService: RecordARecallService,
 ) {
   @PostMapping(value = ["/record-a-recall/{prisonerId}"])
```
yasinmustafa, line 23: this endpoint url would prob be better as 
`/record-a-recall/calculate/{prisonerId}`

but will address that separately

PR #1234, opened 2025-10-08
RCLL-24 Add recalls specific roles to endpoints that recalls uses

yasinmustafa:
Some tweaking of roles to conform with the latest conventions https://dsdmoj.atlassian.net/wiki/spaces/NDSS/pages/4632379405/RFC+4+-+Role+Protecting+API+Endpoints

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/resource/RecordARecallController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/resource/RecordARecallController.kt
@@ -21,7 +21,7 @@ class RecordARecallController(
   private val recordARecallService: RecordARecallService,
 ) {
   @PostMapping(value = ["/record-a-recall/{prisonerId}"])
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/resource/RecordARecallController.kt
```diff
@@ -21,7 +21,7 @@ class RecordARecallController(
   private val recordARecallService: RecordARecallService,
 ) {
   @PostMapping(value = ["/record-a-recall/{prisonerId}"])
-  @PreAuthorize("hasAnyRole('RECORD_A_RECALL', 'ROLE_CALCULATE_RELEASE_DATES__RECALL__CALCULATE__RW')")
+  @PreAuthorize("hasAnyRole('RECORD_A_RECALL', 'CALCULATE_RELEASE_DATES__RECALL__CALCULATE__RW')")
```
yasinmustafa, line 24: you can prefix with ROLE_ (or not), the convention in CRD seems to be without (although it's mixed)

RECORD_A_RECALL role will be eventually removed from here

joelstobart-moj (approved):
LGTM

PR #1238, opened 2025-10-08
CRS-2491: Update things to do logic

james-baker-moj:
Instead of using the analysed sentences and offences we use a more specific set of conditions and also check return to custody date and fine payments. One of the issues with using analysed sentences and offences is that removed sentences were not detected.


ldlharper (approved):
Looks really good 👍

PR #1443, opened 2026-04-09
Library updates

james-baker-moj:
These were raised by renovate but failing as there were some compilation issues as a result.


mg-moj (approved):
lgtm

PR #1444, opened 2026-04-10
CRS-2649: Remove SDS40 Youth Repeal

james-baker-moj:
This is not anticipated to happen now. For now Section 250s should still be excluded from progression model but SDS40 should still apply wherever eligible.


mg-moj (approved):
lgtm

PR #1448, opened 2026-04-14
CRS-2502: Timeline event refactor

james-baker-moj:
Change timeline calculation to be directed by sealed class rather than enum. This allows the removal of nullable types for specific events that need to know additional data such as which SDS legislation or tranche applies. We can also get rid of `currentTimelineCalculationDate` and have the handlers be type safe. 

The enum is maintained as the order of events by type is important but it is now explicit within the enum rather than using the natural order of elements.


mg-moj (approved):
lgtm

PR #1450, opened 2026-04-15
CRS-1160: PED should not be earlier than earliest sentence date

james-baker-moj:
Large amounts of remand could make the PED earlier than the sentence date. If it's earlier than the earliest sentence date in the sentence envelope then default to the earliest sentence date instead.


mg-moj (approved):
lgtm

PR #1451, opened 2026-04-15
crs-2671-ftr56-tranche-boundary

mg-moj:
Implement different sentence to tranche boundary checks for each tranche strategy.

Add unit tests for each legislation.


src/test/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/calculation/StandardConfigurationTests.kt
```diff
@@ -10,6 +11,16 @@ import java.util.stream.Stream
 
 class StandardConfigurationTests : CalculationExampleTests() {
 
+  @Test
+  fun blah() {
```
james-baker-moj: don't think you meant to leave this in

james-baker-moj (approved):
LGTM

PR #1452, opened 2026-04-15
CRS-2672: Filter sentence adjustments with no matching sentence

james-baker-moj:
When a sentenceSequence is specified on an adjustment but there is no matching sentence found in the active sentences then the adjustment does not apply the current sentence envelope.

Added some missing tests to match the coverage provided for prison API adjustments transformation which is no longer used.


mg-moj (approved):
lgtm

PR #1455, opened 2026-04-20
CRS-2652: Recall and Tagged Bail PM tranche tests

james-baker-moj:
Recall and tagged bail associated with sentences before progression model commencement should not be considered in tranche allocation.

This is already handled by the timeline so no further code changes were required.


mg-moj (approved):
lgtm

PR #1474, opened 2026-04-29
CRS-2656: Import Manage Offences model using OpenAPI

james-baker-moj:
Importing MO's model highlighted a difference in the available exclusions with VIOLENT_T3 only existing on the CRDS side.

As this is not a valid exclusion it's now been removed from CRDS.

After reviewing the tests that used VIOLENT_T3 they should have been classed as SEXUAL_T3 according to the original tickets.


mg-moj (approved):
lgtm

PR #1475, opened 2026-04-30
CRS-2656: Use new API for SDS offence details

james-baker-moj:
Now get the PCSC markers and SDS exclusions in a single API call.

Restricted the call to only SDS sentences and now release arrangements are only set on SDS sentences.

Fixed issue with mapping of T3 legislation and removed CRDS side exclusions that don't exist in MO and have never been used.

Also re-enabled some tests that had been commented out.


src/test/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/ReleaseArrangementLookupServiceTest.kt
```diff
@@ -403,71 +449,17 @@ class ReleaseArrangementLookupServiceTest {
     assertThat(withReleaseArrangements[1].sdsReleaseArrangements!!.sdsEarlyReleaseExclusions.firstOrNull()).isEqualTo(SDSEarlyReleaseExclusionType.VIOLENT)
     assertThat(withReleaseArrangements[1].sentenceSequence).isEqualTo(200)
 
-    verify(mockManageOffencesService, times(1)).getPcscMarkersForOffenceCodes(any())
-    verify(mockManageOffencesService).getSdsExclusionsForOffenceCodes(listOf("SX01", "V01"))
-  }
-
-  @Test
-  fun `should not check for sexual or violent if it's SDS plus`() {
```
james-baker-moj, line 411: Now combined call so this is not relevant any more

src/test/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/ReleaseArrangementLookupServiceTest.kt
```diff
@@ -403,71 +449,17 @@ class ReleaseArrangementLookupServiceTest {
...
   @ParameterizedTest
   @CsvSource(
     // sentence date
-    // "DV04001,ADIMP,2022-06-28,7,true",
-    // "DV04001,ADIMP,2022-06-29,7,true",
+    "DV04001,ADIMP,2022-06-28,7,true",
+    "DV04001,ADIMP,2022-06-29,7,true",
     "DV04001,ADIMP,2022-06-27,7,false",
     // sentence length
-    /*"DV04001,ADIMP,2022-06-28,8,true",
```
james-baker-moj, line 470: loads of these tests were commented out for some reason.... think that was an error as I reinstated and they all passed apart from the unsupported sentence types which should now return null rather than false/no exclusion

joelstobart-moj (approved):
LGTM

PR #1485, opened 2026-05-05
CRS-2502: Alt progression model config

james-baker-moj:
An alt progression model config with a commencement date in the past that allows us to test post-commencement scenarios in alt-preprod.

Also removed some dupe config in the regular progression model calc params.


mg-moj (approved):
lgtm

PR #1487, opened 2026-05-06
CRS-2656: Would be SDS+ and Schedule 13 Part 3 exclusions

james-baker-moj:
Schedule 13 Part 3 is only excluded on sentences before progression model commencement. Sentences that would be SDS+ if sentenced today are excluded from progression as they will also have an SDS40 exclusion so will be calculated at 50% which is the same as SDS+ under progression model.


mg-moj (approved):
lgtm

PR #1494, opened 2026-05-09
CRS-2656: Fix unsupported sentence type

james-baker-moj:
Unsupported sentence types are parsed as StandardDeterminateSentence for the purpose of display and analysis for the manual journey. These won't have the release arrangements populated.

Now only require release arrangements to be populated on actual SDS sentences and use default release arrangements for non-SDS offences. Fixed an issue with the transform function test which caused it not to fail for indeterminate sentences.


mg-moj (approved):
lgtm

PR #1495, opened 2026-05-11
CRS-2656: Clean up SentenceAndOffenceWithReleaseArrangements constrctors

james-baker-moj:
Remove the one that takes a prison API sentence and offence. This to clean up some tests that were making use of the back compat code.


joelstobart-moj (approved):
LGTM

PR #1498, opened 2026-05-12
CRS-2720: Reduce concurrency for bulk calc

james-baker-moj:
The default is 10. Set to 5 everywhere except preprod where it's 7. This is due to a production issue caused by an all prisons run maxing out the DBs CPU for an hour.


joelstobart-moj (approved):
LGTM

PR #1502, opened 2026-05-13
CRS-2697: Progression Model ADA handling

james-baker-moj:
Split out SDS early release finalisation for SDS40 and progression model.  Progression model defaulting excludes ADAs in consideration of the date and if defaulted to tranche commencement then ADAs should still be served.


joelstobart-moj (approved):
LGTM

PR #1503, opened 2026-05-13
CRS-2503: Add dates to tranche eligibility tests

james-baker-moj:
Originally we were only validating the tranche eligibility but now these tests check the dates as well and form part of the defaulting test coverage for progression model. Discovered a number of issues with the original test definitions which didn't impact tranche allocation but do impact the dates generated.


mg-moj (approved):
lgtm

PR #1507, opened 2026-05-15
CRS-2684: Operative Sentence Envelope API

james-baker-moj:
Store an operative sentence envelope against CRDS supported calculations. Add an API that will use the stored envelope if there is one and it is the latest calculation or fallback to deriving one against the latest NOMIS calculation.

When deriving from NOMIS we can't provide correct values for the recall and SDS+ flags so these are null.

This endpoint is to support EMS


joelstobart-moj (approved):
LGTM

PR #1509, opened 2026-05-15
CRS-2663: Updated UAL scenarios for progression model

james-baker-moj:
On review of the UAL scenarios there were some that were invalid or missing. Included in that was the addition of post-progression UAL to a defaulted tranche commencement date.


mg-moj (approved):
lgtm

PR #1510, opened 2026-05-18
CRS-2684: Fix swagger docs

james-baker-moj:
There is a bug when using @JsonUnwrapped. Used the workaround from that issue in the spring repo.

Added a test to make sure swagger and Open API specs can be loaded.


build.gradle.kts
```diff
@@ -88,6 +88,7 @@ dependencies {
   testImplementation("io.github.hakky54:logcaptor:2.12.6")
   testImplementation("org.mockito.kotlin:mockito-kotlin")
   testImplementation("uk.gov.justice.service.hmpps:hmpps-kotlin-spring-boot-starter-test:2.2.0")
+  testImplementation("javax.xml.bind:jaxb-api:2.3.1")
```
james-baker-moj, line 91: not sure why this is only required when running in test mode, I've run locally and it's not required but we get a class not found in the tests 🤷

joelstobart-moj (approved):
LGTM

PR #1514, opened 2026-05-19
CRS-2663: Ensure UAL not included in progression defaulting

james-baker-moj:
It was being added post defaulting but should not be considered in the decision to use pre-progression dates or tranche default and then added afterwards only if we default to the tranche date.


mg-moj (approved):
lgtm

PR #1515, opened 2026-05-19
CRS-2684: Update swagger for operative sentence envelope

james-baker-moj:
The default docs made it look like you wouldn't get null from NOMIS for SDS+ and recall flags


joelstobart-moj (approved):
LGTM

PR #1517, opened 2026-05-19
CRS-2730: Use 40% or 50% CRD for HDCED until disabled

james-baker-moj:
Until HDC is disabled for adult sentences it should be calculated using a 40% or 50% CRD regardless of whether progression model is switched on or not.


mg-moj (approved):
lgtm

PR #1520, opened 2026-05-20
CRS-2723: Apply PM defaulting everywhere

james-baker-moj:
When using the sentence level release dates with defaulting the proper rules were not applied for progression model. Centralised this within SDS legislation so that the correct rules are applied in all places.


joelstobart-moj (approved):
LGTM

PR #1525, opened 2026-05-21
CRS-2694: Progression model - remove unconditional release

james-baker-moj:
Sentences under 12 months with an offence date before ORA were eligible for unconditional release (ARD + SED). Under progression model these sentences should now have a conditional release date (CRD + SLED).


joelstobart-moj (approved):
LGTM

PR #1536, opened 2026-06-01
CRS-2736: PM - Remand and Tagged Bail validation

james-baker-moj:
Redirect to manual for remand and tagged bail validation errors where the sentence is extinguished and the prisoner was subject to progression model date. This is because the PM date could cause there to be excess remand/TB but prisons should be able to manually calculate the dates in accordance with the op guidance in these scenarios.

Handling post-progression remand and tagged bail will be done in a subsequent ticket. For now we only need this to work in the operational window.


mg-moj (approved):
lgtm

PR #1539, opened 2026-06-03
CRS-2750: New progression model tranche boundaries and fixes

james-baker-moj:
Including relevant test updates from BAs. Discovered an issue with tranche boundaries not being inclusive for PM and adding a single day to single sentence bookings unnecessarily for all tranche legislations.

Changed the tranche type to `SENTENCE_LENGTH_LESS_THAN` or `SENTENCE_LENGTH_LESS_THAN_OR_EQUAL_TO` to make it explicit it in the config whether the upper bound of the tranche is included or not.

Now uses the same algorithm to calculate sentence duration for both consecutive and single sentences.

Added boundary tests for all tranches.


joelstobart-moj (approved):
LGTM

PR #1540, opened 2026-06-04
CRS-2745: SDS descriptions for UI

james-baker-moj:
Add new API for getting historical sentence and offence data from a calculation that has the same type as when loading for a booking. Previously it was using SentenceAndOffenceWithReleaseArrangements which should be more for internal use.

Add SDS description information to that including SDS 40 exclusion reason, progression model exclusion reason and the way to display SDS plus offences based on sentence type. This removes some of the logic from the UI.

Fix issue where we're not storing the new data model with the correct version number.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -243,3 +243,7 @@ configure<KtlintExtension> {
     }
   }
 }
+val compileKotlin: KotlinCompile by tasks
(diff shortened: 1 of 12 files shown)


joelstobart-moj (approved):
LGTM

PR #1544, opened 2026-06-10
CRS-2657: Affected by progression model flag

james-baker-moj:
Add flag to db to highlight when a calculation was affected by progression model, this is similar to the affected by SDS40 flag.

It only applies when actually calculating dates and should not include instances where the dates do not come from sentences affected by progression model.

Also fixed issue where post-progression sentences could be defaulted to a tranche when selecting the sentence to extract. If that was a short sentence it could end up being an early release. (AC3.1)


joelstobart-moj (approved):
LGTM

PR #1546, opened 2026-06-10
CRS-2742: Additional fields for bulk comparison person overview

james-baker-moj:
Adding the calculation reason of the latest calculation, including further detail for Other and Legislation changes, and also the progression model tranche.


mg-moj (approved):
lgtm

PR #1548, opened 2026-06-11
CRS-2663: Post-Progression Model UAL updates

james-baker-moj:
Post-Progression Model UAL should only be excluded for tranche eligibility but included in tranche defaulting. Originally we also removed it as part of tranche defaulting logic (and then reapplied it).

This also removed the need to keep track of post-progression UAL in the timeline.


mg-moj (approved):
lgtm

PR #1552, opened 2026-06-16
CRS-2732 - Initial Commit for Second Check

FaizJD030:


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/CalculationTransactionalService.kt
```diff
@@ -125,14 +125,16 @@ class CalculationTransactionalService(
     val currentBookingJson = objectToJson(booking, objectMapper)
     val preliminaryBookingJson = calculationRequest.inputData
 
-    if (preliminaryBookingJson.hashCode() != currentBookingJson.hashCode()) {
-      throw PreconditionFailedException("The booking data used for the preliminary calculation has changed")
-    }
+    if (calculationRequest.reasonForCalculation?.isSecondCheck == false) {
```
james-baker-moj: What is the need for this? The booking should be the same between the two calculations regardless surely? is it also skipping validation? I don't think it should fail validation on 2nd check if original calculation was successful

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/HistoricCalculationsService.kt
```diff
@@ -21,10 +21,33 @@ class HistoricCalculationsService(
   @Transactional(readOnly = true)
   fun getHistoricCalculationsForPrisoner(prisonerId: String): List<HistoricCalculation> {
     val calculations = calculationRequestRepository.findAllByPrisonerIdAndCalculationStatus(prisonerId, CONFIRMED.name)
+    val secondChecks = calculations.filter { it.reasonForCalculation?.isSecondCheck == true }
     val nomisCalculations = prisonService.getCalculationsForAPrisonerId(prisonerId)
     val agencyIdToDescriptionMap = prisonService.getAgenciesByType("INST").associateBy { it.agencyId }
     val uniqueUsers: Set<String> = nomisCalculations.mapNotNull { it.calculatedByUserId?.uppercase() }.toSet()
     val userDetails = manageUsersApiClient.getUsersByUsernames(uniqueUsers)
+    val secondCheckReason = "Second Check"
```
james-baker-moj: Can't you use `reasonForCalculation` from CalculationRequest instead of hardcoding it?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/LatestCalculationService.kt
```diff
@@ -43,6 +43,7 @@ class LatestCalculationService(
           bookingId,
           prisonerCalculation,
           nomisReason,
+          latestCrdsCalc.orElse(null)?.calculationType?.name ?: "Unknown",
```
james-baker-moj: Is calculation type necessary here anymore?

james-baker-moj (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/CalculationTransactionalService.kt
FaizJD030: We are skipping validation for second check because when i tested this, it was failing for multiple prisoner in dev who has had their last calculation to be successful. I'm not sure why the validation has failed in those scenarios but i think it should be because of underlying data has changed, so in that case the person doing the second check won't confirm the dates and calculation will stop at the PRELIMINARY status.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/HistoricCalculationsService.kt
FaizJD030: The  calc reason for the second check has long display name for the calc reason select screen and a short one for the calc history. I thought it was going to be ok as we are only adding that reason only for isSecondCheck boolean value as true.

If you think this still need to be amended i could pass the isSecondCheck boolean value to the front end and hard code in the frontend, if that boolean is true. Seems to me that either we do it in the frontend or backend.
If you think otherwise please let me know.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/LatestCalculationService.kt
FaizJD030: Yes we the need the calc type in the frontend to display a banner accordingly, - [Manual, Overriden, Normal].
I wasnt sure if this would be applicable for NOMIS calc so falling back to Unknown in those cases.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/HistoricCalculationsService.kt
FaizJD030: Irrelevant with new code

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/CalculationTransactionalService.kt
FaizJD030: Irrelevant with new code

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/HistoricCalculationsService.kt
```diff
@@ -3,28 +3,52 @@ package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.service
...
+  private val secondCheckRepository: SecondCheckRepository,
 ) {
 
   @Transactional(readOnly = true)
   fun getHistoricCalculationsForPrisoner(prisonerId: String): List<HistoricCalculation> {
     val calculations = calculationRequestRepository.findAllByPrisonerIdAndCalculationStatus(prisonerId, CONFIRMED.name)
+    val secondChecks: List<CalculationRequestSecondCheck> = secondCheckRepository.findAllByPrisonerId(prisonerId)
     val nomisCalculations = prisonService.getCalculationsForAPrisonerId(prisonerId)
     val agencyIdToDescriptionMap = prisonService.getAgenciesByType("INST").associateBy { it.agencyId }
     val uniqueUsers: Set<String> = nomisCalculations.mapNotNull { it.calculatedByUserId?.uppercase() }.toSet()
```
james-baker-moj: I think you need to include the second check usernames here as well otherwise they won't be available unless they've also done a calculation

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/entity/CalculationRequestSecondCheck.kt
```diff
@@ -0,0 +1,31 @@
...
+@Entity
+@Table
+data class CalculationRequestSecondCheck(
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  val id: Long? = null,
+  @NotNull
+  @ManyToOne(optional = false)
+  @JoinColumn(name = "calculationRequestId", nullable = false, updatable = false)
+  val calculationRequest: CalculationRequest,
```
james-baker-moj: It might be better just to map the calculation request id here rather than the whole object. It would make the querying simpler and prevent having to load the calculation request twice when loading the history (once for the calculation and again in the 2nd check)

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/ApprovedDatesService.kt
```diff
@@ -43,6 +43,7 @@ class ApprovedDatesService(
     } else {
       when (latestCalculationRequest.calculationType) {
         CalculationType.MANUAL_DETERMINATE, CalculationType.MANUAL_INDETERMINATE -> unavailable(ApprovedDatesUnavailableReason.PREVIOUS_CALCULATION_MANUAL)
+        CalculationType.SECOND_CHECK -> unavailable(ApprovedDatesUnavailableReason.SECOND_CHECK)
```
james-baker-moj: I don't think you need `CalculationType.SECOND_CHECK` anymore. That would also mean you don't need the ApprovedDatesUnavailableReason reason either

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/CalculationResultEnrichmentService.kt
```diff
@@ -26,8 +30,23 @@ class CalculationResultEnrichmentService(
   private val workingDayService: WorkingDayService,
   private val sdsLegislationConfiguration: SDSLegislationConfiguration,
   private val clock: Clock,
+  private val secondCheckRepository: SecondCheckRepository,
+  private val manageUsersApiClient: ManageUsersApiClient,
 ) {
 
+  fun getSecondCheckDetails(calculationRequestId: Long): SecondCheckDetails? {
```
james-baker-moj: It seems odd to put this in the CalculationResultEnrichmentService. It should probably just be in it's own service. Also shouldn't we be returning all the second checks for a request? Or if we only want the latest in this API then we should call it getLatestSecondCheckDetails.
I would be tempted to just get the raw second check details and then add the usernames where it's used so you can call the manage users API with both calculated and checked by usernames in one call. This will result in multiple calls to manage users.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/LatestCalculationService.kt
james-baker-moj: I think calculation type is no longer being used for 2nd check so we can clean this up now

src/test/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/CalculationResultEnrichmentServiceTest.kt
```diff
@@ -1200,6 +1212,33 @@ class CalculationResultEnrichmentServiceTest {
...
+    whenever(secondCheckRepository.findLatestByCalculationRequestId(any())).thenReturn(null)
+    whenever(manageUsersApiClient.getUserByUsername("username")).thenReturn(UserDetails("username", "User Name"))
+
+    val results: SecondCheckDetails? = calculationResultEnrichmentService().getSecondCheckDetails(calculationRequest().id())
+    assertThat(results).isNull()
+  }
+
+  @Test
+  fun `SecondCheck found success path`() {
+    val calcRequest2 = secondCheckRecord().copy(
```
james-baker-moj: what's the 2 for here?

james-baker-moj (changes requested):

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/HistoricCalculationsService.kt
```diff
@@ -3,28 +3,52 @@ package uk.gov.justice.digital.hmpps.calculatereleasedatesapi.service
...
     val userDetails = manageUsersApiClient.getUsersByUsernames(uniqueUsers)
+    val secondCheckCalculations = secondChecks.map { secondCheck: CalculationRequestSecondCheck ->
```
james-baker-moj: I thought we were not going to model these as calculations here and instead add them as a field on the relevant calculations?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/repository/SecondCheckRepository.kt
```diff
@@ -0,0 +1,15 @@
...
+import org.springframework.stereotype.Repository
+import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.entity.CalculationRequestSecondCheck
+
+@Repository
+interface SecondCheckRepository : JpaRepository<CalculationRequestSecondCheck, Long> {
+  fun findAllByPrisonerId(prisonerId: String): List<CalculationRequestSecondCheck>
+
+  @Query("SELECT c FROM CalculationRequestSecondCheck c WHERE c.calculationRequestId = :calculationRequestId ORDER BY c.checkedAt DESC LIMIT 1")
+  fun findLatestByCalculationRequestId(@Param("calculationRequestId") calculationRequestId: Long): CalculationRequestSecondCheck?
+}
```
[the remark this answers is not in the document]
FaizJD030: The existing code works as expected.

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/HistoricCalculationsService.kt
FaizJD030: Addressed

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/entity/CalculationRequestSecondCheck.kt
FaizJD030: Addressed

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/ApprovedDatesService.kt
FaizJD030: Removed

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/CalculationResultEnrichmentService.kt
FaizJD030: Moved

src/test/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/CalculationResultEnrichmentServiceTest.kt
FaizJD030: Amended

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/service/HistoricCalculationsService.kt
FaizJD030: Moved

james-baker-moj (dismissed):
LGTM

james-baker-moj (approved):
LGTM

PR #1555, opened 2026-06-18
crs-2761-hdced-repeal-bug

mg-moj:
Always exclude HDCED if sentence is affected by progression legislation


james-baker-moj (approved):
LGTM

PR #1557, opened 2026-06-22
CRS-2764: Apply defaulting to saved SDS40 calc

james-baker-moj:
The "before tranche allocation" calculation for SDS40 was being overwritten by the one for Progression Model and the PM "before tranche allocation" calculation didn't have SDS40 defaulting rules applied.

This meant that where a pre-SDS40 date had been retained as part of SDS40 it was no longer retained for PM and therefore the date could be later than intended. This ensures pre-SDS40 dates can still be retained with PM defaulting.


mg-moj (approved):
lgtm

PR #1561, opened 2026-06-25
CRS-2775: ERS30 defaulting not considering UAL

james-baker-moj:
When the ERSED gets defaulted to the sentence date UAL was not being considered in the ERS30 commencement check leading to a potential early release as the ERS30 commencement date was before the sentence date + UAL.


yasinmustafa (approved):
lgtm

PR #1562, opened 2026-06-26
Try -XX:+UseContainerSupport flag

james-baker-moj:
This should allow the JVM to be sized appropriately for the container. This will hopefully resolve OOM issues in dev.


joelstobart-moj (approved):
LGTM

PR #1565, opened 2026-06-30
CRS-2737: Enable Progression Model for operational window

james-baker-moj:
Enabling progression model in prod for the operational window where HDC will still be available if the date is before progression commencement but all other calculations will be performed using progression model release parameters. This is to allow recalculations to take place before commencement and the first tranche of releases.

Also switching dev and alt-dev around.


joelstobart-moj (approved):
LGTM

PR #1566, opened 2026-07-01
CRS-2779: Update error messages for extinguished sentence

james-baker-moj:
The first of many update messages that route users to the appropriate DPS service.

New category as the default validation category instructs users to go to NOMIS then reload the page.

Add content type to the validation message to allow for richer content in the messages. Existing plain text content retained as the current formatting changes new lines to <br/> tags.


joelstobart-moj (approved):
LGTM

PR #1567, opened 2026-07-02
CRS-2781: Ignore TUSED diff in bulk calc

james-baker-moj:
Due to the PSS repeal we no longer generate TUSEDs. Many prisoners are showing up in bulk calc with a difference because their previous calculation produced a TUSED and now doesn't.  We're now ignoring the TUSED as a difference.


joelstobart-moj (approved):
LGTM

PR #1575, opened 2026-07-10
CRS-2791: Ignore adjustments have matching sentence validation for IR

james-baker-moj:
Identify remand needs to be able to run the calculation despite validation errors that might be caused by remand. This is so that users are able to recalculate remand correctly using the IR tool.


mg-moj (approved):
lgtm

PR #1578, opened 2026-07-15
RASS-2451 - Additional Role permission added

dmaggio-jd:
Additional permission added to /calculation/{prisonerId}/latest to support access for HMPPS-Court-Cases-Release-Dates

LatestCalculationIntTest.kt lightly refactored to keep DRY test range of expected permissions


mg-moj (approved):
lgtm

PR #1580, opened 2026-07-16
CRS-2796: Fix PED defaulting issue

james-baker-moj:
When applying SDS40 defaulting for the "before tranche calculation" for PM we were only passing in the released sentence groups. This works when applied at the end of the timeline but not at the PM tranche dates. Include the current sentence group in the defaulting (which is the same as we do for the calculation itself).


mg-moj (approved):
lgtm

PR #1581, opened 2026-07-17
crs-2798-validate-adjustment-dates

mg-moj:
Validate adjustment from date is before to date.

Resolve deprecation and compiler warnings.


src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/validator/AdjustmentsBeforeCalculationValidator.kt
```diff
@@ -27,13 +29,26 @@ class AdjustmentsBeforeCalculationValidator(private val validationUtilities: Val
...
     addAll(validateAllRemandHasFromAndToDates(adjustments))
     addAll(validateBookingAdjustment(adjustments.bookingAdjustments))
-    addAll(
-      validateRemandOverlappingRemand(
-        adjustments.sentenceAdjustments
-          .filter { it.type == SentenceAdjustmentType.REMAND && it.fromDate != null && it.toDate != null }
-          .map { LocalDateRange.of(it.fromDate!!, it.toDate!!) },
-      ),
-    )
+    val (validAdjustment, invalidAdjustments) = adjustments.sentenceAdjustments
```
james-baker-moj: maybe `[in]validRemand` instead of adjustment?

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/external/SentenceAdjustment.kt
```diff
@@ -22,3 +22,13 @@ enum class SentenceAdjustmentType {
   TIME_SPENT_IN_CUSTODY_ABROAD,
   TIME_SPENT_AS_AN_APPEAL_APPLICANT,
 }
+
+fun SentenceAdjustment.getHumanReadableAdjustmentType(): String = when (this.type) {
+  SentenceAdjustmentType.RECALL_SENTENCE_REMAND -> "Recall Sentence Remand"
+  SentenceAdjustmentType.RECALL_SENTENCE_TAGGED_BAIL -> "Recall Sentence Tagged Bail"
```
james-baker-moj: I think these also just need to be Remand and Tagged Bail as the recall-ness of it is not represented in adjustments only CRDS internals

src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/validator/AdjustmentsBeforeCalculationValidator.kt
```diff
@@ -27,13 +29,26 @@ class AdjustmentsBeforeCalculationValidator(private val validationUtilities: Val
...
+      .filter { it.type == SentenceAdjustmentType.REMAND && it.fromDate != null && it.toDate != null }
+      .partition { it.fromDate!! <= it.toDate }
+
+    if (invalidAdjustments.isNotEmpty()) {
```
james-baker-moj: You also need to do all this under `validateAdjustmentsBeforeCalculation` which takes `adjustments: List<AdjustmentDto>`. Both methods are here while the feature toggle exists but I think we could start to drop the old `BookingAndSentenceAdjustments` versions of things soon. Validation only performed on the latest data rather than old calculations so no need to retain backwards compatibility any more.

src/test/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/validation/validator/AdjustmentsBeforeCalculationValidatorTest.kt
```diff
@@ -0,0 +1,177 @@
...
+import uk.gov.justice.digital.hmpps.calculatereleasedatesapi.validation.ValidationUtilities
+import java.time.LocalDate
+
+class AdjustmentsBeforeCalculationValidatorTest {
+
+  private val validator = AdjustmentsBeforeCalculationValidator(ValidationUtilities(), ADJUSTMENTS_UI_URL)
+
+  @Test
+  fun `returns invalid date range when remand from date is after to date`() {
+    val adjustments = BookingAndSentenceAdjustments(
```
james-baker-moj: Probably best to make the tests be `AdjustmentDto`s version instead (or both) as that's what's actually enabled in reality.

james-baker-moj (approved):
LGTM

PR #1585, opened 2026-07-21
crs-2800-progression-t1-manual-journey

mg-moj:
Trigger manual journey validation where calculation result is affected by progression model and assigned to Tranche 1. Only apply where feature toggle is enabled.


james-baker-moj (dismissed):
LGTM

james-baker-moj (approved):

PR #1586, opened 2026-07-21
CRS-2689: Post-progression consec sentences

james-baker-moj:
Tests for sentences imposed consecutively post-progression. Fixed small issue with affected by progression model where there is  SDS40 defaulting required but not PM defaulting.


mg-moj (approved):
lgtm

PR #1587, opened 2026-07-21
CRS-2802: Separate HDC clean stop from progression commencement

james-baker-moj:
Turns out it was already separated so just renamed the constant to make it clear.


mg-moj (approved):
lg

PR #1589, opened 2026-07-28
CRS-2787: Prisoner calculation overview API

james-baker-moj:
API for the prisoner overview which includes a detailed calculation as well as a number of historical calculation summaries based on the request and the total number of calculations on a prisoner.


joelstobart-moj (approved):
LGTM

PR #1593, opened 2026-07-31
CRS-2810: Post-ERS30 sentences should not default

james-baker-moj:
Sentences that are post-ERS30 or, in the case of consecutive sentences, contain both pre and post-ERS30 commencement sentences should not be subject to defaulting to the ERS30 commencement date even if the date is earlier.

There are further changes to be made in relation to ERSED and progression model which will be addressed in a future ticket and ERSED calculation has been temporarily disabled for now. Fixes coming with CRS-2689.


mg-moj (approved):
lgtm

PR #1595, opened 2026-08-05
CRS-2813: New exclusions for progression model

james-baker-moj:
Initially we will route to manual until the calculation rules can be assessed. Both parsing the new exclusion from MO and routing to manual are managed via feature toggle.


mg-moj (approved):
lgtm

PR #1597, opened 2026-08-06
CRS-2816: Reword PM excluded offence display name

james-baker-moj:
So it shows "Excluded offence" on check sentence and offence info page.


mg-moj (approved):
all good

PR #1599, opened 2026-08-07
CRS-2818: New and update APIs for probation

james-baker-moj:
Add an API that can lookup whether a prisoner has an offence excluded from progression model. Also added the flag to operative sentence envelope.


mg-moj (approved):
lgtm

PR #1607, opened 2026-08-13
CRS-2812: ERS30 bug and snapshot refactor

james-baker-moj:
Separate out the responsibility of the tranche timeline handler into specific events for allocating, recalculating and taking snapshots.

Implement a snapshot for ERS30 commencement and then use the later of that snapshot or the current timeline value when extracting ERSED.

This is to handle the scenario where a sentence is imposed consecutively after ERS30 commencement. While the consecutive chain at the point of ERS30 commencement will have a shorter overall sentence length it can still produce a later ERSED as it would have been calculated using ERS50. We need to maintain that date as that is the ERSED that would have applied before the consecutive chain was extended but they must not have been released under ERS as a subsequent sentence was imposed. If we were to take the earlier of the two dates it would have been before they could've actually been released under ERS. This keeps it historically accurate and is the approach directed to us by policy.


mg-moj (approved):
lgtm

PR #1609, opened 2026-08-17
CRS-2812: Feature toggle for ERS30 snapshot

james-baker-moj:
There may be issues with this setup so feature toggling the snapshot and turning it off for now except in the tests that were added/updated in the ticket


mg-moj (approved):
lgtm

